PR #6, opened 2019-06-05
Cleanup

chuckha:
This adds some godoc, has a few minor improvements to language and runs go fmt across a few files.
This also updates dep.


ncdc:
LGTM, thank you!

PR #17, opened 2019-06-12
Eliminate global logging

### LINKED ISSUE #11: Logging improvements
Logging is global right now and should not be. If we ever care about actual output of this for, say, automation, we will want to test output and using global loggers makes this very hard.

Logging output should be uniform. If we can avoid klog/glog that would be great because they encourage global logging and are harder for automated log ingestion. Structured logging (like logrus which is what is currently being used) providers more utility and configuration.

ncdc:
Remove all global logging calls to logrus.*.

Pull in go-logr as our logging interface.

Add a logrus implementation for go-logr.

Fixes #11

### DIFF
--- a/pkg/logging/logrus_logr.go
+++ b/pkg/logging/logrus_logr.go
@@ -1,0 +1,70 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package logging
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"github.com/sirupsen/logrus"
+)
+
+type logrusAdapter struct {
+	l *logrus.Entry
+}
+
+func NewLogrusLoggerAdapter(l *logrus.Logger) logr.Logger {
+	return NewLogrusEntryAdapter(logrus.NewEntry(l))
+}
+
+func NewLogrusEntryAdapter(e *logrus.Entry) logr.Logger {
+	return &logrusAdapter{
+		l: e,
+	}
+}
+
+func (l *logrusAdapter) withFields(keysAndValues []interface{}) *logrus.Entry {
+	if len(keysAndValues)%2 != 0 {
+		panic("programmer error - must have an even number of keysAndValues")
+	}
+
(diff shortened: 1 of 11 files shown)


pkg/logging/logrus_logr.go
```diff
@@ -0,0 +1,71 @@
...
+
+func NewLogrusEntryAdapter(e *logrus.Entry) logr.Logger {
+	return &logrusAdapter{
+		l: e,
+	}
+}
+
+func (l *logrusAdapter) withFields(keysAndValues []interface{}) *logrus.Entry {
+	if len(keysAndValues)%2 != 0 {
+		fmt.Printf("ANDY %d %#v\n", len(keysAndValues), keysAndValues)
```
chuckha: accidentally a line

pkg/logging/logrus_logr.go
```diff
@@ -0,0 +1,71 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package logging
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
```
chuckha: Oh right, another reason logr is hard to work with is because writing a shim is non trivial. totally fine here, just a followup from our chat yesterday

pkg/logging/logrus_logr.go
```diff
@@ -0,0 +1,71 @@
...
+	// Don't try to distinguish between info, debug, trace as far as V() levels go
+	return true
+}
+
+func (l *logrusAdapter) Error(err error, msg string, keysAndValues ...interface{}) {
+	l.withFields(keysAndValues).WithError(err).Error(msg)
+}
+
+func (l *logrusAdapter) V(level int) logr.InfoLogger {
+	panic("not implemented")
```
chuckha: nice

pkg/logging/logrus_logr.go
ncdc: Thanks for catching 😄

ncdc:
@chuckha got rid of my debugging line

chuckha:
Thanks!

PR #21, opened 2019-06-13
Refactor validations

chuckha:
* Uses cobra to mark flags as required instead of a validate method
* Validate now validates datat including valid scope
* Refactor scope and provide a better error message

### DIFF
--- a/pkg/upgrade/config.go
+++ b/pkg/upgrade/config.go
@@ -10,6 +10,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	ControlPlaneScope      = "control-plane"
+	MachineDeploymentScope = "machine-deployment"
+)
+
 // Config contains all the configurations necessary to upgrade a Kubernetes cluster.
 type Config struct {
 	ManagementCluster ManagementClusterConfig `json:"managementCluster"`
@@ -33,6 +38,10 @@ type TargetClusterConfig struct {
 	UpgradeScope      string        `json:"scope"`
 }
 
+func (t *TargetClusterConfig) UpgradeScopes() []string {
+	return []string{ControlPlaneScope, MachineDeploymentScope}
+}
+
 // KeyPairConfig is something
 type KeyPairConfig struct {
 	SecretRef    string `json:"secretRef,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/upgrade/config.go
```diff
@@ -10,6 +10,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	ControlPlane      = "control-plane"
```
ncdc: I'm not sure these technically need to live here vs `main.go`. It's probably a toss-up, although I doubt we'd move the `upgradeCluster` logic from `main.go` into the `pkg` code.

Wherever they live, I think it's nice to have a common word to group together related constants, so either ScopeControlPlane & ScopeMachineDeployment or ControlPlaceScope and MachineDeploymentScope. WDYT?

pkg/upgrade/config.go
```diff
@@ -53,20 +62,21 @@ type ImageUpdateConfig struct {
...
 	if config.TargetCluster.CAKeyPair.ClusterField != "" && config.TargetCluster.CAKeyPair.SecretRef != "" {
 		return errors.New("only one of key pair cluster field and secret ref may be set")
 	}
 	if config.TargetCluster.CAKeyPair.ClusterField == "" && config.TargetCluster.CAKeyPair.SecretRef == "" {
 		return errors.New("one of key pair cluster field or secret ref must be set")
 	}
-	if config.KubernetesVersion == "" {
-		return errors.New("kubernetes version is required")
+	validUpgradeScope := false
+	for _, scope := range config.TargetCluster.UpgradeScopes() {
```
ncdc: Would you want to use kube's `sets.String` for this?

pkg/upgrade/config.go
chuckha: I don't really car tbh. The other thing I've seen is to do like, `type Scope string` but I tend to find that more onerous than beneficial.

Will update to add scope to the suffix.

pkg/upgrade/config.go
chuckha: I usually don't bother dragging in deps for something this simple, seems like overkill since I'm guessing sets support way more than just this one thing I need. How about if we need this in a few more places we can bring it in?

pkg/upgrade/machine_deployment.go
```diff
@@ -61,15 +61,15 @@ func (u *MachineDeploymentUpgrader) listMachineDeployments() (*clusterapiv1alpha
 func (u *MachineDeploymentUpgrader) upgradeMachineDeployments(list *clusterapiv1alpha1.MachineDeploymentList) error {
 	for _, machineDeployment := range list.Items {
 		if err := u.updateMachineDeployment(&machineDeployment); err != nil {
-			u.log.Error(err, "Failed to create new MachineDeployment", "namespace", machineDeployment.Namespace, "name", machineDeployment.Name)
+			u.log.Error(err, "Failed to create new MachineDeploymentScope", "namespace", machineDeployment.Namespace, "name", machineDeployment.Name)
```
ncdc: Revert please 😄

pkg/upgrade/machine_deployment.go
```diff
@@ -61,15 +61,15 @@ func (u *MachineDeploymentUpgrader) listMachineDeployments() (*clusterapiv1alpha
...
 			return err
 		}
 	}
 	return nil
 }
 
 func (u *MachineDeploymentUpgrader) updateMachineDeployment(machineDeployment *clusterapiv1alpha1.MachineDeployment) error {
-	u.log.Info("Updating MachineDeployment", "namespace", machineDeployment.Namespace, "name", machineDeployment.Name)
+	u.log.Info("Updating MachineDeploymentScope", "namespace", machineDeployment.Namespace, "name", machineDeployment.Name)
```
ncdc: Revert please 😄

chuckha: lol I thought Goland was smarter than that

PR #23, opened 2019-06-13
Support modifying image id when upgrading.

### LINKED ISSUE #9: Support updating image id
We haven't yet fleshed out `config.machineUpdates.image.id/field`. As a user, I want to be able to specify the desired new OS image ID (e.g. ami-xyz), and a specific field in my providerSpec that needs to be updated with this value.

ncdc:
Fixes #9 

### DIFF
--- a/pkg/upgrade/machine_image_test.go
+++ b/pkg/upgrade/machine_image_test.go
@@ -1,0 +1,30 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package upgrade
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	clusterapiv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+)
+
+func TestUpdateMachineSpecImage(t *testing.T) {
+	machine := &clusterapiv1alpha1.Machine{
+		Spec: clusterapiv1alpha1.MachineSpec{
+			ProviderSpec: clusterapiv1alpha1.ProviderSpec{
+				Value: &runtime.RawExtension{
+					Raw: []byte(`{"foo":{"bar": "baz"}}`),
+				},
+			},
+		},
+	}
+
+	err := updateMachineSpecImage(&machine.Spec, "providerSpec.value.foo.bar", "abcd1234")
+	require.NoError(t, err)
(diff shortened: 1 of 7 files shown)


pkg/upgrade/machine_image_test.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package upgrade
```
chuckha: I think we should only be testing the public interface of the packages. What do you think about moving this to package `upgrade_test` and testing Upgrade()?

I would argue a few points on this. 1) if we test internal functions then we're testing an implementation that shouldn't matter to the public function. If the public function does what it's supposed to do, then the internal implementation doesn't matter and we can assume it works. And 2) if it's too difficult to test only the public interface then we need to restructure the code to make this code path testable. 

I'd actually say delete this test and work towards making `updateMachineSpecImage` testable through calling Upgrade.

ncdc: I'd say I generally agree with your suggestion. This is actually something I was chatting about with @vincepri earlier today. The problem with Upgrade() is that it performs a lot of actions and involves a lot of different external resources.

I have yet to find a good way to unit test a large major function. If you start substituting in mocks all over the place, the unit test code gets really messy, especially for table tests. And if you add in a new mockable behavior, you then have to modify most if not all of the unit tests for your large major function.

Do you have any pointers?

chuckha: More tightly scoped interfaces are what I would do here. For example instead of the machineClusterAPIClient interface I'd define a local interface that scopes the interface(s) down to just `create, delete, list, get`. 

If we introduce new behavior on an interface then it makes sense that we have to modify our tests. I've found it to be quite manageable.

I don't think we should use gomocks though, that library creates unnecessary overhead and tests waay too much stuff. 

I can write up some code and see if you like it?

ncdc: Totally 👎 on go mocks, agreed!

If you have a few minutes to put together what you're thinking, I'd love to see it. No need to do it all - if you can flesh out an outline/pseudo, that's fine.

ncdc:
@chuckha do you want me to remove the unit test and merge this as-is for now, and plan to add better tests later (as part of refactoring to make it more testable)?

ncdc:
As discussed on Zoom - going to merge this as-is for now. We can move the test case code around to another place in the future if that makes more sense.

ncdc:
As discussed on Zoom - going to merge this as-is for now. We can move the test case code around to another place in the future if that makes more sense.

PR #25, opened 2019-06-13
[DO NOT MERGE] refactor for easier testing

chuckha:
This is how I would scope down the interfaces.

This outlines at least one reason I don't like client go. I really wish the `.Pods(namespace)` (and friends) returned a struct that stored the namespace internally instead of an interface.

### DIFF
--- a/pkg/upgrade/base.go
+++ b/pkg/upgrade/base.go
@@ -17,12 +17,34 @@ import (
 	clusterapiv1alpha1client "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1"
 )
 
+type ConfigMapNamespacer interface {
+	ConfigMaps(string) NamespacedConfigMapClient
+}
+type NamespacedConfigMapClient interface {
+	Get(string, metav1.GetOptions) (*v1.ConfigMap, error)
+	Create(*v1.ConfigMap) (*v1.ConfigMap, error)
+}
+type PodNamespacer interface {
+	Pods(string) NamespacedPodsClient
+}
+type NamespacedPodsClient interface {
+	Get(string, metav1.GetOptions) (*v1.Pod, error)
+	List(options metav1.ListOptions) (*v1.PodList, error)
+}
+
+type NodeClient interface {
+	List(options metav1.ListOptions) (*v1.NodeList, error)
+}
+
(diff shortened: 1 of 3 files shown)


pkg/upgrade/base.go
```diff
@@ -17,12 +17,34 @@ import (
 	clusterapiv1alpha1client "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1"
 )
 
+type ConfigMapNamespacer interface {
```
chuckha: The thing I really like about this is that our dependencies are listed out very clearly here. I know this app only needs to Get and Create config maps, List nodes, Get and List pods. It does not need watches or anything fancy.

If these change that means there is change in functionality (probably) and I'd fully expect the tests to have to change to reflect that.

pkg/upgrade/control_plane_test.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"testing"
+
+	"github.com/go-logr/logr"
+	"github.com/vmware/cluster-api-upgrade-tool/pkg/upgrade"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type log struct {}
```
chuckha: This is one reason I don't like using logr as our interface for logging, we have to implement the whole interface and I don't believe we use the entire thing

pkg/upgrade/control_plane.go
```diff
@@ -33,6 +33,17 @@ type ControlPlaneUpgrader struct {
 	oldNodeToEtcdMember map[string]string
 }
 
+func NewTestableControlPlaneUpgrader(log logr.Logger, cmn ConfigMapNamespacer, pn PodNamespacer, nc NodeClient) *ControlPlaneUpgrader {
```
chuckha: I'd assert that newBase is too smart and the logic should be kept somewhere else. probably some client factory would be sufficient.

chuckha:
one last comment: I'm not sure how to deal with the pod.Exec needing the entire Kubernetes interface, but I'm *sure* there's some work around we can do for testing there without having to mock the entire k8s interface.

chuckha:
reviewed on zoom.

PR #27, opened 2019-06-14
Refactor how secret lookups work.

### LINKED ISSUE #24: Support secret ref to kubeconfig
We can skip generating a kubeconfig if we store the kubeconfig in the secrets and reference that instead.

chuckha:
Fixes #24

Adds ability to pass a secret ref to a kubeconfig in addition to passing
a ref to a ca cert/key or a path in the provider spec to a ca cert/key pair.


pkg/upgrade/kubeconfig.go
```diff
@@ -0,0 +1,231 @@
...
+}
+type kubeconfigClient interface {
+	Write(clientcmdapi.Config) ([]byte, error)
+	RESTConfigFromKubeConfig([]byte) (*restclient.Config, error)
+	NewKubeconfig(string, string, *x509.Certificate, *rsa.PrivateKey) (*clientcmdapi.Config, error)
+	DecodeCertPEM([]byte) (*x509.Certificate, error)
+	DecodePrivateKeyPEM([]byte) (*rsa.PrivateKey, error)
+}
+
+type kubeconfig struct{}
```
chuckha: need this to wrap global functions to satisfy the interface

pkg/upgrade/kubeconfig_test.go
```diff
@@ -0,0 +1,92 @@
...
+				},
+			}
+			kc := &kubeconfig{
+				certPEM:      &tc.cert,
+				privateKey:   &tc.key,
+				clientConfig: &tc.clientConfig,
+				restConfig:   &tc.restConfig,
+			}
+			cluster := v1alpha1.Cluster{}
+			rc, err := upgrade.GetRestConfig(sc, base64.StdEncoding, kc, &cluster, "", upgrade.KeyPairConfig{})
```
chuckha: this is about as complex as I want unit test logic to be. (no logic at all)

pkg/upgrade/kubeconfig.go
```diff
@@ -0,0 +1,231 @@
...
+	"k8s.io/client-go/rest"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/services/certificates" // RIP
+	clusterapiv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+)
+
+type secretsClient interface {
+	Get(string, metav1.GetOptions) (*v1.Secret, error)
```
ncdc: Would you mind adding names for all the parameters in all these interfaces?

pkg/upgrade/kubeconfig.go
```diff
@@ -0,0 +1,231 @@
...
+
+type keyPairGetter struct {
+	secretsClient      secretsClient
+	decoder            decoder
+	kubeconfigClient   kubeconfigClient
+	unstructuredClient unstructuredFunctions
+}
+
+// GetRestConfig figures out how to build a rest.Config from one of three possible ways based on the KeyPairConfig passed in.
+func GetRestConfig(secretsClient secretsClient, decoder decoder, kubeconfigClient kubeconfigClient, cluster *clusterapiv1alpha1.Cluster, targetClusterAPIEndpoint string, config KeyPairConfig) (*rest.Config, error) {
```
ncdc: Before I go any further on the review, I have one main question: now that we have 2 different approaches to getting a rest config, do you think it's still worth it to consolidate all this logic into a single function? I'm finding the signature (6 parameters) to be unwieldy. Do you think there's a way to make this cleaner?

chuckha:
As discussed on zoom, I'll refactor this into three functions and push the figure-out-which-one-to-call logic up to the caller.

chuckha:
superseded by #32

PR #29, opened 2019-06-14
restructure config

chuckha:
The test is mostly a skeleton to add to if we find it necessary to test the Boolean logic.

EDIT: The bulk of this is now test code boilerplate.
EDIT2: this conflicts with the other open PR, I don't mind fixing the conflicts whenever either of these merge.

### DIFF
--- a/pkg/upgrade/base.go
+++ b/pkg/upgrade/base.go
@@ -4,8 +4,6 @@
 package upgrade
 
 import (
-	"fmt"
-
 	"github.com/blang/semver"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


pkg/upgrade/base.go
```diff
@@ -58,18 +56,6 @@ func newBase(log logr.Logger, config Config) (*base, error) {
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
-	clusterEndPoint := config.TargetCluster.TargetApiEndpoint
```
chuckha: This logic goes away. The user should specify the entire API endpoint & port. putting this burden on the user gives this function fewer things to do while making it more extensible. It now supports any port you like, not just a hardcoded 6443.

pkg/upgrade/config.go
```diff
@@ -44,8 +44,36 @@ func (t *TargetClusterConfig) UpgradeScopes() []string {
...
-	SecretRef    string `json:"secretRef,omitempty"`
-	ClusterField string `json:"clusterField,omitempty"`
+	SecretRef           string `json:"secretRef,omitempty"`
+	ClusterField        string `json:"clusterField,omitempty"`
+	KubeconfigSecretRef string `json:"kubeconfigSecretRef,omitempty"`
+
+	// APIEndpoint is a URL to a kube-apiserver.
+	// This entry is used only if SecretRef or ClusterField is set.
+	// APIEndpoint is ignored if KubeconfigSecretRef is set.
+	APIEndpoint string `json:"apiEndpoint"`
```
chuckha: I moved this into the KeyPairConfig object because this is the object it is used with/required for.

There is also an argument for keeping it in TargetClusterConfig and moving the following validation function onto the TargetClusterClient. I don't really care either way, this just made the most sense to me at the time of writing.

ncdc:
Looks pretty good to me

chuckha:
@ncdc 100% coverage of Validate as requested!

ncdc:
LGTM. Ready to merge?

chuckha:
Yep, ready when you are

PR #32, opened 2019-06-17
Secret lookup

chuckha:
79.5% coverage of key_pair.go. Can't test error cases without doing more dependency injection, but I think this is a nice compromise.


pkg/upgrade/key_pair.go
```diff
@@ -8,42 +8,55 @@ import (
...
-	if config.ClusterField != "" {
-		return getEmbeddedKeyPair(cluster, config.ClusterField)
+	kc, ok := secret.Data[KubeconfigSecretKey]
+	if !ok {
+		return nil, errors.Errorf("item 'kubeconfig' not found in secret %q", name)
 	}
-
-	return k.getSecretRefKeyPair(cluster.Namespace, config.SecretRef)
+	out := make([]byte, base64.StdEncoding.DecodedLen(len(kc)))
+	n, err := base64.StdEncoding.Decode(out, kc)
```
chuckha: this actually doesn't need to be decoded, the client does that automatically which really confused the heck out of me.

pkg/upgrade/key_pair.go
```diff
@@ -8,42 +8,55 @@ import (
...
+type keyPair struct {
 	Cert []byte `json:"cert"`
 	Key  []byte `json:"key"`
 }
 
-type keyPairGetter struct {
-	secretClient corev1client.SecretsGetter
+// secrets has already been scoped to namespace, so, .Secrets("my-namespace") returns one of these.
+type secrets interface {
+	Get(string, metav1.GetOptions) (*v1.Secret, error)
```
ncdc: If you wouldn't mind, please name the parameters

pkg/upgrade/key_pair.go
ncdc: Does this mean you need to remove this `Decode` call?

pkg/upgrade/key_pair.go
```diff
@@ -8,42 +8,55 @@ import (
...
+	out = out[:n]
+	return clientcmd.RESTConfigFromKubeConfig(out)
 }
 
-func getEmbeddedKeyPair(cluster *clusterapiv1alpha1.Cluster, path string) (*KeyPair, error) {
-	pathParts := strings.Split(path, ".")
+// NewRestConfigFromClusterField returns a rest.Config configured with the CA key pair found in the cluster's
+// object in the fieldpath specified. For example, "spec.providerSpec.value.caKeyPair" traverses the cluster
+// object going through each '.' delimited field.
+func NewRestConfigFromClusterField(cluster *clusterv1alpha1.Cluster, fieldPath, apiEndpoint string) (*rest.Config, error) {
```
ncdc: Should we call this `NewRestConfigFromCAClusterField`?

pkg/upgrade/key_pair.go
```diff
@@ -8,42 +8,55 @@ import (
...
 	}
-
-	return k.getSecretRefKeyPair(cluster.Namespace, config.SecretRef)
+	out := make([]byte, base64.StdEncoding.DecodedLen(len(kc)))
+	n, err := base64.StdEncoding.Decode(out, kc)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	out = out[:n]
+	return clientcmd.RESTConfigFromKubeConfig(out)
```
ncdc: We should wrap the error here

pkg/upgrade/key_pair.go
```diff
@@ -8,42 +8,55 @@ import (
...
 	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/services/certificates"
-	clusterapiv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+	clusterv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 )
 
-// KeyPair contains a cert and key.
-type KeyPair struct {
+// KubeconfigSecretKey is the key where the kubeconfig is stored in the secret.
+const KubeconfigSecretKey = "kubeconfig"
```
ncdc: unexport?

pkg/upgrade/key_pair_test.go
```diff
@@ -0,0 +1,187 @@
...
+	secret := &secrets{
+		secret: &v1.Secret{
+			Data: map[string][]byte{
+				"cert": []byte(caCertificate),
+				"key":  []byte(caPrivateKey),
+			},
+		},
+	}
+	config, err := upgrade.NewRestConfigFromCASecretRef(secret, "name", "clustername", "https://example.com:6443")
+	if err != nil {
```
ncdc: Feel free to pull in testify and do `require.NoError(t, err)`, etc

chuckha: that seems like a big dependency for a very small syntax change. Is there any other reason you think we should be pulling that in?

pkg/upgrade/key_pair.go
chuckha: yes, that is correct

pkg/upgrade/key_pair_test.go
ncdc: I quite like testify for unit testing. If we pull it in, we can use it in all our tests.

chuckha: I don't think we need it for this statement as `err != nil` is an extremely common pattern in go. If it provides value in some other way in the future I'll be happy to pull it in

ncdc: Fair enough

pkg/upgrade/key_pair.go
chuckha: yeah, that's fine. The purpose of this is to make it obvious what the secret key is and that purpose is still served with an unexported const 👍

pkg/upgrade/key_pair.go
chuckha: yeah, no worries, that definitely improves readability

pkg/upgrade/key_pair.go
chuckha: Actual result:

* Do not store secrets that you encode with base64. `kubectl` automatically encodes the data. If you *do* store b64 secrets, you will need to read them with `kubectl get secret ... | base64 --decode | base64 --decode`.
* Assume that providers are storing secrets that are *not* b64 encoded.

pkg/upgrade/key_pair.go
chuckha: sure, seems a little more clear

chuckha:
updated! ptal whenever you have time @ncdc

ncdc:
LGTM!

PR #36, opened 2019-06-19
Integration test for upgrade control plane

chuckha:
Code is a bit rough but this successfully runs a control-plane upgrade on a simple cluster (1 worker 1 control plane) 

Requires `capd` binary, found here
https://github.com/chuckha/cluster-api-provider-docker

Not yet a blocking test, but this seems like a good place to share work now that it actually runs through the real code with real services.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -1,0 +1,282 @@
+package main
+
+import (
+	"bytes"
+	"encoding/base64"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/vmware/cluster-api-upgrade-tool/pkg/upgrade"
+)
+
+func TestUpgradeScenario(t *testing.T) {
+	// spin up management cluster
+	setupManagementCluster(t)
+
+	clusterName := "my-cluster"
+	namespace := "default"
+
+	// Create a test cluster to upgrade
+	createTestCluster(t, clusterName, namespace, "v1.14.1")
+
+	// wait for kubeconfig to show up
+	fmt.Println("Waiting up to 5 minutes for a the kubeconfig secret to appear")
+	if err := waitForKubeconfigSecret(clusterName, namespace); err != nil {
+		panic(err)
+	}
+
+	fmt.Println("Waiting for up to 5 minutes for etcd to be ready")
+	if err := waitForEtcdReady(clusterName, namespace); err != nil {
+		panic(fmt.Sprintf("%+v\n", err))
+	}
+
+	fmt.Println("ready for testing!")
+	// upgrade from 1.14.1 to 1.14.2
+
+	// management cluster kubeconfig
+	path, err := kind("get", "kubeconfig-path", "--name", "kind")
+	if err != nil {
+		panic(err)
+	}
+
+	path = bytes.TrimSpace(path)
+	cfg := upgrade.Config{
+		KubernetesVersion: "v1.14.2",
+		ManagementCluster: upgrade.ManagementClusterConfig{
+			Kubeconfig: string(path),
+		},
+		TargetCluster: upgrade.TargetClusterConfig{
+			Name:         clusterName,
+			Namespace:    namespace,
+			UpgradeScope: upgrade.ControlPlaneScope,
+			CAKeyPair: upgrade.KeyPairConfig{
+				KubeconfigSecretRef: secretName(clusterName),
+			},
+		},
+	}
+	if err := upgradeCluster(cfg); err != nil {
+		t.Fatalf("%+v", err)
+	}
+}
+
+func waitForKubeconfigSecret(clusterName, namespace string) error {
+	timer := time.NewTimer(5 * time.Minute)
+	ticker := time.Tick(10 * time.Second)
+	for {
+		select {
+		case <-timer.C:
+			return errors.New("timed out waiting for kubeconfig to appear")
+		case <-ticker:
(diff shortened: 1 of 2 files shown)


main_test.go
```diff
@@ -0,0 +1,282 @@
...
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	defer os.Remove(kubeconf.Name())
+	if err := ioutil.WriteFile(kubeconf.Name(), out, os.FileMode(0644)); err != nil {
+		return errors.WithStack(err)
+	}
+	timeout := time.NewTimer(5 * time.Minute)
+	ticker := time.NewTicker(5 * time.Second)
+	for {
```
ncdc: use `wait.PollImmediate`?

main_test.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	if err := upgradeCluster(cfg); err != nil {
+		t.Fatalf("%+v", err)
+	}
+}
+
+func waitForKubeconfigSecret(clusterName, namespace string) error {
+	timer := time.NewTimer(5 * time.Minute)
+	ticker := time.Tick(10 * time.Second)
+	for {
```
ncdc: use `wait.PollImmediate`?

main_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/vmware/cluster-api-upgrade-tool/pkg/upgrade"
+)
+
+func TestUpgradeScenario(t *testing.T) {
+	// spin up management cluster
+	setupManagementCluster(t)
+
+	clusterName := "my-cluster"
```
ncdc: Should we use a common prefix and generate a random suffix for this & the namespace?

main_test.go
```diff
@@ -0,0 +1,282 @@
...
+	namespace := "default"
+
+	// Create a test cluster to upgrade
+	createTestCluster(t, clusterName, namespace, "v1.14.1")
+
+	// wait for kubeconfig to show up
+	fmt.Println("Waiting up to 5 minutes for a the kubeconfig secret to appear")
+	if err := waitForKubeconfigSecret(clusterName, namespace); err != nil {
+		panic(err)
```
ncdc: should we `t.Fatal*` instead, here & elsewhere?

main_test.go
```diff
@@ -0,0 +1,282 @@
...
+	setupManagementCluster(t)
+
+	clusterName := "my-cluster"
+	namespace := "default"
+
+	// Create a test cluster to upgrade
+	createTestCluster(t, clusterName, namespace, "v1.14.1")
+
+	// wait for kubeconfig to show up
+	fmt.Println("Waiting up to 5 minutes for a the kubeconfig secret to appear")
```
ncdc: Should we use `t.Log*` instead of `fmt.Print*`?

main_test.go
```diff
@@ -0,0 +1,282 @@
...
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/vmware/cluster-api-upgrade-tool/pkg/upgrade"
+)
+
+func TestUpgradeScenario(t *testing.T) {
+	// spin up management cluster
+	setupManagementCluster(t)
```
ncdc: Do we need a `defer` to tear this down?

main_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+	clusterName := "my-cluster"
+	namespace := "default"
+
+	// Create a test cluster to upgrade
+	createTestCluster(t, clusterName, namespace, "v1.14.1")
```
ncdc: Do we need a `defer` to do cleanup?

main_test.go
```diff
@@ -0,0 +1,282 @@
...
+	if e, ok := err.(*exec.ExitError); ok {
+		return string(e.Stderr)
+	}
+	return err.Error()
+}
+
+func setupManagementCluster(t *testing.T) {
+	t.Helper()
+	// if a cluster named kind already exists then we assume we're good to go:
+	// TODO: this isn't very good because we have to clean up the management cluster or use unique machines etc.
```
ncdc: +1 to uniqueness

main_test.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	return pipeToKubectlApply(bytes.NewReader(out))
+}
+
+func capdctl(args ...string) ([]byte, error) {
+	cmd := exec.Command("capdctl", args...)
+	return cmd.Output()
+}
+
+// TODO this input is not right...
```
ncdc: Yeah, something like kubectl().ForCluster(c).WithReader(r).Run("apply", "-f", "-") might be nice?

chuckha:
I 100% agree with all of your feedback. However, I'm curious if you would mind if I made incremental improvements or do you see these as baseline code standards that no code should be merged without?

ncdc:
I'm totally fine w/incremental. Just leave TODOs where needed?

chuckha:
Perf, lemme update these with TODOs. Thank you for the feedback, I also like your fluent command api :D

chuckha:
Added a bunch of TODOs at the top

main_test.go
```diff
@@ -0,0 +1,292 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/vmware/cluster-api-upgrade-tool/pkg/upgrade"
+)
+
+// Code improvements:
+// TODO: replace fmt.Println with t.Log
+// TODO: instead of panicking fatal the test case
+// TODO: Look into wait.PollImmediate instead of this home-grown solution, dep might not be worth it though, depends on how much code it is
```
ncdc: We're already using this dep

ncdc:
Ran this locally. It worked! 🎉

Issue #1, opened 2019-05-31
Idempotency during upgrade


kaushik229:
Where possible we should attempt to introspect state from the underlying objects (such as the defined Kubernetes version). Where that is not possible, it might be a good idea to track the state using annotations (or a CRD) rather than storing state locally within the tool.

We need to add support for making sure cluster upgrades are idempotent.
In case of failure, we must start from the state where previous run of upgrade failed.

chuckha:
Annotations along with the algorithm above seem to be a reasonable first approach, but it sounded like we might need a rethink here, any additional concerns I should know about @ncdc?

### Annotations as state machine
How is this not a library yet? It's out of scope right now but I've seen this idea come up a number of times and a library would be nice. I'm pretty ok starting out with the proposed algorithm above, or something very similar.

#### States
* Pending => upgrade has not started
* Started (Upgrading?) => upgrade is in the process of happening
* Finished => upgrade has finished (this state will be deleted almost immediately)
* No state => No known upgrade requests
* Error => Error was encountered during an upgrade. May require human intervention.

ncdc:
I'm not sure it makes the most sense to use an annotation on the `Cluster` that is an enum `UpgradeAll | UpgradeControlPlane | UpgradeWorkers`. For upgrading workers (machine deployments), that is achieved by changing the kubelet version in the machine deployment spec. Therefore I don't think we need `UpgradeAll` or `UpgradeWorkers`. That leaves `UpgradeControlPlane`, but it doesn't contain enough information. I think I'd rather see something that indicates "desired control plane version", and our upgrade tool can evaluate that value against the version for each control plane machine (and possibly other annotation metadata) to determine what to do. That's my initial thought. Will add more soon.

detiber:
@ncdc if we want a "no-op" type upgrade where we upgrade the cluster without upgrading the k8s version, we will need to trigger off of something more than just the k8s version.

ncdc:
True, thanks for pointing that out. Will need to figure out how to define that as part of the "possibly other annotation metadata" 😄

ncdc:
For same-version control plane machine replacements, we could have the upgrade tool generate a unique "control plane revision/id" and store that in an annotation. If a control plane machine does not also have that same annotation & value, we'd consider that machine stale and in need of replacement. We'd then create a new control plane machine (presumably taking advantage of a newer OS image) and make sure we annotate the machine with this "control plane revision/id".

[...]

PR #42, opened 2019-06-20
Adds upgrade ID flag and writes annotations to upgraded nodes

chuckha:
Related to #1 

### DIFF
--- a/main.go
+++ b/main.go
@@ -76,6 +76,9 @@ func main() {
 	root.Flags().StringVar(&upgradeConfig.MachineUpdates.Image.Field, "image-field",
 		"", "The image identifier field in provider manifests (optional)")
 
+	root.Flags().StringVar(&upgradeConfig.UpgradeID, "upgrade-id", "",
+		"Unique identifier used to resume a partial upgrade (optional)")
+
 	if err := root.Execute(); err != nil {
 		// Print a stack trace, if possible. We may end up with the error message printed twice,
 		// but the stack trace can be invaluable, so we'll accept this for the time being.
@@ -108,5 +111,7 @@ func upgradeCluster(config upgrade.Config) error {
 		return err
 	}
 
+	infoMessage := fmt.Sprintf("Rerun with `--upgrade-id=%s` if this upgrade fails midway and you want to retry", config.UpgradeID)
+	log.Info(infoMessage)
 	return upgrader.Upgrade()
 }
--- a/pkg/upgrade/control_plane.go
+++ b/pkg/upgrade/control_plane.go
@@ -28,6 +28,9 @@ const (
 	etcdCACertFile = "/etc/kubernetes/pki/etcd/ca.crt"
 	etcdCertFile   = "/etc/kubernetes/pki/etcd/peer.crt"
 	etcdKeyFile    = "/etc/kubernetes/pki/etcd/peer.key"
+
+	// UpgradeIDAnnotationKey is the annotation key for this tool's upgrade-id
+	UpgradeIDAnnotationKey = "upgrade-id"
 )
 
 type ControlPlaneUpgrader struct {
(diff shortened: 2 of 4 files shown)


main.go
```diff
@@ -76,6 +76,8 @@ func main() {
 	root.Flags().StringVar(&upgradeConfig.MachineUpdates.Image.Field, "image-field",
 		"", "The image identifier field in provider manifests (optional)")
 
+	root.Flags().StringVar(&upgradeConfig.UpgradeID, "upgrade-id", "", "Supply an the identifier to resume an partial upgrade (optional)")
```
ncdc: Maybe remove `Supply`?

chuckha: I'll reword it

pkg/upgrade/control_plane.go
```diff
@@ -261,6 +261,12 @@ func (u *ControlPlaneUpgrader) updateMachines(machines *clusterapiv1alpha1.Machi
 
 	// TODO add more error logs on failure conditions
 	for _, machine := range machines.Items {
+		annotations := machine.GetAnnotations()
+		// Skip any machine that already has the annotation we're looking for
+		if val, ok := annotations["upgrade-id"]; ok && val == u.upgradeID {
```
ncdc: Let's move the name of the annotation to a constant. We should also namespace it in some way.

pkg/upgrade/control_plane.go
```diff
@@ -322,10 +328,20 @@ func (u *ControlPlaneUpgrader) updateMachines(machines *clusterapiv1alpha1.Machi
...
+		if err := u.applyAnnotation(&machine); err != nil {
+			return err
+		}
 	}
 	return nil
 }
 
+func (u *ControlPlaneUpgrader) applyAnnotation(m *clusterapiv1alpha1.Machine) error {
+	m.Annotations["upgrade-id"] = u.upgradeID
+	_, err := u.managementClusterAPIClient.Machines(u.clusterNamespace).Update(m)
```
ncdc: How about Patch instead of Update?

main.go
```diff
@@ -76,6 +77,9 @@ func main() {
 	root.Flags().StringVar(&upgradeConfig.MachineUpdates.Image.Field, "image-field",
 		"", "The image identifier field in provider manifests (optional)")
 
+	root.Flags().StringVar(&upgradeConfig.UpgradeID, "upgrade-id", fmt.Sprintf("%d", time.Now().Unix()),
```
ncdc: I'm wondering if having the defaulting logic for `upgradeConfig.UpgradeID` here is the right place for it? This is the front-end CLI and it might not always be used. How would you feel about putting the defaulting into `newBase` (for now) if the field is unset by the time we get there?

pkg/upgrade/control_plane.go
ncdc: Can add a TODO for this for now

main.go
chuckha: Yep, that makes sense, putting logic in the default value is not great

pkg/upgrade/control_plane.go
chuckha: `com.vmware.cluster-api-upgrade-tool.upgrade-id` ? i have always been bad at these java style name spacing 😆

ncdc: Eh, let's just do a constant for now - ok to keep it as `upgrade-id`. If we feel like it, we can file an issue after this is merged to adjust the label key later.

pkg/upgrade/control_plane.go
```diff
@@ -26,6 +26,9 @@ const (
 	etcdCACertFile = "/etc/kubernetes/pki/etcd/ca.crt"
 	etcdCertFile   = "/etc/kubernetes/pki/etcd/peer.crt"
 	etcdKeyFile    = "/etc/kubernetes/pki/etcd/peer.key"
+
+	// UpgradeIDAnnotationKey is the annotation key used to mark vmware's cluster-api upgrade-tool's upgrade-id
```
ncdc:
```suggestion
	// UpgradeIDAnnotationKey is the annotation key for this tool's upgrade-id
```

chuckha: fair

ncdc:
LGTM - want to squash?

chuckha:
i can squash to make a decent commit message, one sec

chuckha:
@ncdc ok, done!

ncdc:
LGTM, thanks!

PR #44, opened 2019-06-21
setup a test for multicp upgrade

chuckha:
This test runs the upgrade against a 3 node control plane. it takes about 10 minutes to complete on an xl instance in aws.

`go test -count=1 -timeout 20m -v -run TestMultipleControlPlaneUpgradeScenario .`

code cleanups after idempotency work gets tested.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ncdc:
LGTM but as we discussed on Slack, we could use `jsonpath` output using the `range/end` syntax instead of custom columns (to avoid having to skip the header line).

chuckha:
Not sure if you missed https://github.com/vmware/cluster-api-upgrade-tool/pull/44/files#diff-53618cae32e23ae47e1c35f10da92654R160 or if you're asking me to remove that line in favor of jsonpath. Can you clarify what you'd like done and why? I don't see a benefit to deleting the `--no-header` and converting to jsonpath, but I want to hear your take

ncdc:
Recording from Slack: my bad - missed the `--no-headers`.

Ready for me to merge?

chuckha:
yes! go for it

PR #47, opened 2019-06-25
Extract an object responsible for creating machines

chuckha:
ControlPlaneUpgrader was doing a lot. One thing it was doing
was creating machines and waiting for the backing node to get to a certain
state.

In order to properly test idempotency we need to be able to stop the Upgrade
at some point. This refactor starts to allow us to have full control
over what actions the Upgrader is performing.

Note: This is to give you an idea of the direction I'd like to head. I will add unit tests to this object if you like it, otherwise we can talk about it!


pkg/upgrade/machine_creator.go
```diff
@@ -0,0 +1,283 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package upgrade
+
+import (
+	"fmt"
+	"github.com/blang/semver"
```
chuckha: ehhhh i have some imports thing i have to figure out on my new computer. i lost settings (intentionally) and just haven't set this one up yet. plz ignore.

pkg/upgrade/machine_creator.go
```diff
@@ -0,0 +1,283 @@
...
+	if n.MachineOptions.ImageField != "" && n.MachineOptions.ImageID != "" {
+		if err := updateMachineSpecImage(&newMachine.Spec, n.MachineOptions.ImageField, n.MachineOptions.ImageID); err != nil {
+			return nil, nil, err
+		}
+	}
+
+	newMachine.Spec.Versions.ControlPlane = n.MachineOptions.DesiredVersion.String()
+	newMachine.Spec.Versions.Kubelet = n.MachineOptions.DesiredVersion.String()
+
+	n.log.Info("Creating new machine", "name", newMachine.GetName())
```
ncdc: Personal preference, but I tend to access the fields directly (`.Name`) vs calling the accessor functions (`GetName()`, etc.). Do you have thoughts on why 1 might be better than the other?

pkg/upgrade/machine_creator.go
```diff
@@ -0,0 +1,283 @@
...
+func (n *MachineCreator) waitForMatchingNode(providerID string, timeout time.Duration) (*v1.Node, error) {
+	n.log.Info("Waiting for node", "provider-id", providerID)
+	var matchingNode v1.Node
+	err := wait.PollImmediate(5*time.Second, timeout, func() (bool, error) {
+		nodes, err := n.nodeLister.List(metav1.ListOptions{})
+		if err != nil {
+			return false, err
+		}
+		for _, node := range nodes.Items {
+			if node.Spec.ProviderID == providerID {
```
ncdc: Now that the provider id code is in the CAPI `release-0.1` branch, we'll want to switch to using it. This line here won't work for e.g. AWS because the node's provider id will be something like `aws://us-east-1/i-xxxxx` and the machine's provider id doesn't contain the region.

pkg/upgrade/machine_creator.go
chuckha: nothing that matters. I tend to favor getters in case there is logic in them that i don't want to miss, but I know in this case there isn't. I'll default to directly accessing the object in future

pkg/upgrade/machine_creator.go
chuckha: oh ok, could you open an issue with the details? Seems out of scope for this PR

pkg/upgrade/machine_creator.go
```diff
@@ -0,0 +1,283 @@
...
+type MachineOptions struct {
+	ImageID string
+	ImageField string
+	DesiredVersion semver.Version
+}
+
+// MachineCreator is responsible for creating a new machine.
+// It can optionally wait for the associated node to become ready.
+type MachineCreator struct {
+	WaitForProviderID bool
```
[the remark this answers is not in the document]
chuckha: capital letters are more appealing to me -- i will make them hidden though as it doesn't affect anything

pkg/upgrade/machine_creator.go
```diff
@@ -0,0 +1,283 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+	clusterapiv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type podGetter interface {
+	Get(string, metav1.GetOptions) (*v1.Pod, error)
```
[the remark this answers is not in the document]
chuckha: I'm not at the point where I think that's worth it, but I have more experience using the static client. I think we should open an issue so we can talk about it there. I'd say it's out of scope for this PR

pkg/upgrade/machine_creator.go
ncdc: Sure, but this is a regression if you leave it like this. For now, you'll need to run `kubernetes.ParseProviderID(node.Spec.ProviderID)` and compare it with `providerID`. Make sense?

chuckha: oh, hmm, we are doing that, the parsed provider ID comes back from `waitForProviderID` https://github.com/vmware/cluster-api-upgrade-tool/pull/47/files#diff-d11660ea136b9301cfeac8702fc73ca1R157

ncdc: You have to parse `node.Spec.ProviderID` here, which is missing

chuckha: Ah. yes. totally. my fault, excellent code reading skills

chuckha: Instead of parsing *both* I think it makes more sense to parse neither and do less work.

since `providerID` is already parsed and I have to parse `node.Spec.ProviderID`, this could instead parse neither  -- or is there something i'm missing from ProviderIDs?

ncdc: The k/k cloud provider, at least for AWS, includes the region in the provider id. CAPA does not include the region. A straight up string comparison will not work. The best solution is to use the `Equals()` function from https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/controller/noderefutil/providerid.go.

chuckha: ok, thanks for explaining~! got it

pkg/upgrade/control_plane.go
```diff
@@ -264,8 +227,27 @@ func (u *ControlPlaneUpgrader) updateMachines(machines *clusterapiv1alpha1.Machi
...
+		ImageID: u.imageID,
+		ImageField: u.imageField,
+		DesiredVersion: u.desiredVersion,
+	}
+
 	// TODO add more error logs on failure conditions
 	for _, machine := range machines.Items {
+		// This can move outside the for loop if all machines will be in the same namespace.
+		machineCreator := NewMachineCreator(
+			WithMachineCreator(u.managementClusterAPIClient.Machines(u.clusterNamespace)),
```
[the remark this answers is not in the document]
chuckha: yeah, i'm not sure what to rename this right now, I'm adding a TODO. Agreed it looks super weird.

pkg/upgrade/control_plane.go
```diff
@@ -264,8 +227,27 @@ func (u *ControlPlaneUpgrader) updateMachines(machines *clusterapiv1alpha1.Machi
...
+			WithMachineGetter(u.managementClusterAPIClient.Machines(machine.Namespace)),
+			WithNodeLister(u.targetKubernetesClient.CoreV1().Nodes()),
+			WithPodGetter(u.targetKubernetesClient.CoreV1().Pods("kube-system")),
+			WithMachineOptions(mo),
+			WithLogger(u.log.WithName("machine-creator")),
+			WaitForProviderID,
```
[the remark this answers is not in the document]
chuckha: Done. Defaulted to true and added options to set it to false if desired

ncdc:
LGTM (can iterate more in future PRs as needed). Ready for merge? Have you run the main upgrade integration test w/these changes?

pkg/upgrade/control_plane.go
```diff
@@ -334,18 +284,34 @@ func (u *ControlPlaneUpgrader) updateMachines(machines *clusterapiv1alpha1.Machi
...
 	return nil
 }
 
 func (u *ControlPlaneUpgrader) applyAnnotation(m *clusterapiv1alpha1.Machine) error {
-	m.Annotations[UpgradeIDAnnotationKey] = u.upgradeID
-	// TODO(chuckha): use Patch instead of Update
-	_, err := u.managementClusterAPIClient.Machines(u.clusterNamespace).Update(m)
-	return errors.WithStack(err)
+	maxTries := 5
+	for i := 0; i < maxTries; i++ {
```
chuckha: I went with this to avoid importing controller runtime just for generating an object patch. I've not had much fun importing controller runtime in other (capi/capa) projects.

ncdc: We don't need controller-runtime for a patch. We can use client-go and json-patch.

chuckha: I haven't seen client-go export any feature that allows me to create a patch, can you point me to a function that client-go uses to generate patches or the specific json-patch library you're referring to. There seem to be a number of them.

ncdc: https://github.com/heptio/velero/blob/dd96aa76dbd1e3af36df6e60fab985d45b1d9d92/pkg/controller/download_request_controller.go#L226-L248

chuckha: oh. Do you mean generate a patch by hand that is a json-patch and then use client-go to call .Patch? That's a totally reasonable request

ncdc: Yes, you marshal before & after to json, then pass both jsons to `jsonpatch.CreateMergePatch`, then use client-go's Patch() call.

chuckha: yep, that non-k8s library was the missing piece.

chuckha: Updated!

pkg/upgrade/control_plane.go
```diff
@@ -334,18 +287,31 @@ func (u *ControlPlaneUpgrader) updateMachines(machines *clusterapiv1alpha1.Machi
...
 	}
 	return nil
 }
 
 func (u *ControlPlaneUpgrader) applyAnnotation(m *clusterapiv1alpha1.Machine) error {
+	original, err := json.Marshal(m)
+	if err != nil {
+		return errors.WithStack(err)
+	}
 	m.Annotations[UpgradeIDAnnotationKey] = u.upgradeID
```
ncdc: Do we need a nil check on the annotations map, with initialization if it's nil?

chuckha: no, ObjectMeta is not a pointer and neither is Annotations. But maps are kinda pointers by themselves so I ran this test: https://play.golang.org/p/wD2cRIV_vEY

We don't need one, but if you'd feel more comfortable with one, I can add it, easy enough.

ncdc: https://play.golang.org/p/dP6_hRTMbMY 😄

chuckha: hah, ok ok, i'll add a nil check

ncdc:
Ready to merge?

chuckha:
yep, integration just passed

chuckha:
(it passed before, but this was with the nil check 😅 )

PR #57, opened 2019-07-03
Edit kubeadm ConfigMap for control plane upgrades

### LINKED ISSUE #56: Control plane upgrades run the wrong images in the new machines
`kubeadm join` determines the container image tag for the static pod manifests from the kubeadm configmap in the cluster. We must modify this when upgrading, before we create any new machines.

ncdc:
When a new control plane node joins the cluster, it gets the version to
use as the image tag for the static pod manifests from the
kubeadm-config ConfigMap in the kube-system namespace. We must change
the ClusterConfiguration.kubernetesVersion to match the desired version,
then upload the updated ConfigMap before proceeding to create any new
machines.

Fixes #56

### DIFF
--- a/pkg/upgrade/control_plane.go
+++ b/pkg/upgrade/control_plane.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"time"
 
+	"sigs.k8s.io/yaml"
+
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/blang/semver"
@@ -88,6 +90,10 @@ func (u *ControlPlaneUpgrader) Upgrade() error {
 	}
 
 	if err := u.UpdateProviderIDsToNodes(); err != nil {
+		return err
+	}
+
+	if err := u.updateAndUploadKubeadmKubernetesVersion(); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/upgrade/control_plane.go
```diff
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"time"
 
+	"sigs.k8s.io/yaml"
+
```
chuckha: nit: we should probably pull in those hack/verify scripts that @neolit123 has been posting around to our new sig cluster lifecycle projects. They are actually really awesome to have!

pkg/upgrade/control_plane.go
```diff
@@ -518,3 +524,41 @@ func (u *ControlPlaneUpgrader) etcdctlForPod(ctx context.Context, pod *v1.Pod, a
 
 	return stdout, stderr, err
 }
+
+func (u *ControlPlaneUpgrader) updateAndUploadKubeadmKubernetesVersion() error {
```
chuckha: Do you have any interest in pulling this kind of work out into its own struct, maybe a struct responsible for modifying the kubeadm config? I was thinking if we doin't want to run `kubeadm reset` on the actual nodes themselves we will have to modify the kubeadm config anyway. Might be good to have that struct and tests in place when we get around to fixing that bug as well.

but i could see it as being useful

I don't mind this though, I think this is clear enough

ncdc: I'd prefer to merge this as-is and open issue(s) as needed for future work, if that's ok with you?

pkg/upgrade/control_plane.go
ncdc: Add an issue please?

pkg/upgrade/control_plane.go
chuckha: definitely. Merging.

PR #61, opened 2019-07-05
Adds a machine deployment upgrade test

chuckha:
~~This test takes a long time because cluster-api takes a long time to rollout machine deployment updates~~

~~Please see this for more information: https://github.com/kubernetes-sigs/cluster-api/issues/1127~~

EDIT: we fixed the bug.

There are a bunch of improvements with this code:

* `kind` is no longer a dependency of the integration tests. They will be even easier to run.
* There is some new documentation to show how to run these tests
* Allow for customization of CAPI and CAPD images which is useful because both are under heavy development
* Some general cleanups, more TODOs but we'll get there.


main_test.go
```diff
@@ -16,15 +17,36 @@ import (
...
 // TODO: CLEANUP CODE, probably use a defer. Not sure what kind of cleanup to do. Do we want to try to reuse the management cluster?
 //       should each test get its own unique cluster and then have a single global teardown of the management cluster?
 // TODO: Fixup the kubectl command to be nicer to use. It has a super weird signature right now. Suggested from review:
 //       kubectl().ForCluster(c).WithReader(r).Run("apply", "-f", "-") might be nice!
+// TODO: create a namespace. right now everything is default since that ns already exists.
+
+const (
+	workerNodeType       = "worker"
+	controlPlaneNodeType = "control-plane"
+	// use latest release-0.1 after the noderef bug gets fixed
```
ncdc: Can switch to v0.1.6 now

main_test.go
```diff
@@ -73,6 +94,59 @@ func TestMultipleControlPlaneUpgradeScenario(t *testing.T) {
...
+
+func TestMachineDeployment(t *testing.T) {
+	setupManagementCluster(t)
+	clusterName := "machine-deployment"
+	namespace := "default"
+	numberOfControlPlaneNodes := 1
+	workerReplicas := 2
+
+	if err := createTestControlPlane(clusterName, namespace, "v1.14.1", numberOfControlPlaneNodes); err != nil {
+		t.Fatal(fmt.Sprintf("%+v", err))
```
ncdc: Do you want to use handleErr here?

main_test.go
```diff
@@ -73,6 +94,59 @@ func TestMultipleControlPlaneUpgradeScenario(t *testing.T) {
...
+	}
+	if err := createMachineDeployment(t, clusterName, namespace, "v1.14.1", workerReplicas); err != nil {
+		t.Fatal(fmt.Sprintf("%+v", err))
```
ncdc: Do you want to use handleErr here?

main_test.go
```diff
@@ -84,29 +158,28 @@ func TestUpgradeScenario(t *testing.T) {
...
+	if err := createTestControlPlane(clusterName, namespace, "v1.14.1", numberOfControlPlaneNodes); err != nil {
+		t.Fatal(handleErr(err))
+	}
+	createWorkers(t, clusterName, namespace, "v1.14.1", 1)
 
 	// wait for kubeconfig to show up
 	fmt.Println("Waiting up to 5 minutes for a the kubeconfig secret to appear")
 	if err := waitForKubeconfigSecret(clusterName, namespace); err != nil {
-		panic(err)
+		t.Fatal(fmt.Sprintf("%+v", err))
```
ncdc: Do you want to use handleErr here?

main_test.go
```diff
@@ -84,29 +158,28 @@ func TestUpgradeScenario(t *testing.T) {
...
 	}
 
 	fmt.Println("Waiting for up to 5 minutes for etcd to be ready")
 	if err := waitForEtcdReady(clusterName, namespace, numberOfControlPlaneNodes); err != nil {
-		panic(fmt.Sprintf("%+v\n", err))
+		t.Fatal(fmt.Sprintf("%+v", err))
```
ncdc: Do you want to use handleErr here?

main_test.go
```diff
@@ -290,22 +422,16 @@ func setupWorkerObject(clusterName, namespace, version string) error {
...
-}
-
-// workerName generates the base worker node name for a given cluster
-func workerName(clusterName string) string {
-	return fmt.Sprintf("%s-worker", clusterName)
+	return fmt.Sprintf("%s-%s%d", clusterName, kind, number)
 }
 
 // secretName generates the name of the kubeconfig secret stored on the management cluster
 func secretName(clusterName string) string {
```
ncdc: We can replace this with CAPI's `remote.KubeConfigSecretName`

main_test.go
```diff
@@ -143,6 +216,52 @@ func waitForKubeconfigSecret(clusterName, namespace string) error {
...
+
+	readyRe := regexp.MustCompile(`\s+Ready`)
+	timeout := time.NewTimer(5 * time.Minute)
+	ticker := time.NewTicker(5 * time.Second)
+	for {
+		select {
+		case <-timeout.C:
+			return errors.New("timed out waiting for ready nodes")
+		case <-ticker.C:
+			cmd := exec.Command("kubectl",
```
ncdc: Doesn't need to be in this PR, but would you want to switch to using client-go instead of shelling out to kubectl?

main_test.go
```diff
@@ -143,6 +216,52 @@ func waitForKubeconfigSecret(clusterName, namespace string) error {
...
+				"--no-headers",
+			)
+			out, err := cmd.Output()
+			if err != nil {
+				return errors.WithStack(err)
+			}
+			lines := bytes.Split(bytes.TrimSpace(out), []byte("\n"))
+			idCount := 0
+			for _, l := range lines {
+				if readyRe.Match((l)) {
```
ncdc: Extra parens?

main_test.go
```diff
@@ -143,6 +216,52 @@ func waitForKubeconfigSecret(clusterName, namespace string) error {
...
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	defer os.Remove(kubeconf.Name())
+	if err := ioutil.WriteFile(kubeconf.Name(), data, os.FileMode(0644)); err != nil {
+		return errors.WithStack(err)
+	}
+
+	readyRe := regexp.MustCompile(`\s+Ready`)
+	timeout := time.NewTimer(5 * time.Minute)
```
ncdc: I'd still recommend using `wait.PollImmediate` but if you want to keep this, we should defer stopping the timer & ticker.

main_test.go
```diff
@@ -237,50 +363,56 @@ func waitForEtcdReady(clusterName, namespace string, expected int) error {
...
-		t.Fatal(err)
+		return err
 	}
-	if err := setupWorkerObject(clusterName, namespace, version); err != nil {
-		t.Fatal(handleErr(err))
+	return nil
+}
+
+func createWorkers(t *testing.T, clusterName, namespace, version string, numberOfWorkers int) {
+	t.Helper()
```
ncdc: I find this makes it harder to figure out what broke if it came from in here - wdyt?

chuckha: yeah, i agree with this

main_test.go
chuckha: Agreed! It is finally time!

main_test.go
chuckha: yeah, agreed. i'll file an issue

main_test.go
chuckha: 🦅 👀

ncdc:
@chuckha LGTM - have you run the tests manually to confirm everything works?

chuckha:
ok, i'm happy with it, ran one and fixed a nil map from the accidental push to master

ncdc:
LGTM, thanks. Anything we missed we can do in a follow-up.

PR #62, opened 2019-07-08
Use Clusterctl's provider ID parsing

liztio:


ncdc:
Hold please

ncdc:
I was thinking we could remove this repo's `ParseProviderID` function, since we're vendoring the one from CAPI

ncdc:
@liztio would you be willing to make this change ^^^^ ?

pkg/internal/kubernetes/provider_id_test.go
```diff
@@ -0,0 +1,54 @@
...
+		{
+			name:  "amazon with extras",
+			value: "amazon://some/values/machinename",
+			id:    "machinename",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			id, err := ParseProviderID(test.value)
```
chuckha: I read these as 2 different test cases. Trying to shove error conditions into a test case that is only checking successful responses makes the logic in this for loop too complex. These t.Run functions should be as small as possible with as little logic as possible.

The two test cases I see are `TestProviderIDReturnsErrorWithBadInput`, `TestProviderIDReturnsProviderID`

pkg/internal/kubernetes/provider_id.go
```diff
@@ -4,19 +4,17 @@
...
 import (
-	"strings"
-
 	"github.com/pkg/errors"
+	"sigs.k8s.io/cluster-api/pkg/controller/noderefutil"
 )
 
 // ParseProviderID is a convenience function to pull out the provider ID
 // @TODO add switch case based on different providers
 func ParseProviderID(providerID string) (string, error) {
```
chuckha: not 100% sure we need to keep this function, but it's probably ok for now

liztio:
@ncdc updated

pkg/upgrade/base.go
```diff
@@ -132,12 +133,16 @@ func (u *base) UpdateProviderIDsToNodes() error {
...
+		if err == nil {
+			id = providerID.ID()
+		} else {
+			u.log.Error(err, "failed to parse provider id", "id", node.Spec.ProviderID)
 			// unable to parse provider ID with whitelist of provider ID formats. Use original provider ID
-			providerID = node.Spec.ProviderID
+			id = node.Spec.ProviderID
 		}
-		pairs[providerID] = &node
+		pairs[id] = &node
```
ncdc: Can we use `providerID` as the key instead of `id`? That would allow us to compare on both provider (e.g. `aws`) and id, instead of just id.

pkg/upgrade/control_plane.go
```diff
@@ -272,12 +273,17 @@ func (u *ControlPlaneUpgrader) updateMachines(machines *clusterapiv1alpha1.Machi
...
-		originalProviderID, err := kubernetes.ParseProviderID(*machine.Spec.ProviderID)
+		originalProviderID, err := noderefutil.NewProviderID(*machine.Spec.ProviderID)
 		if err != nil {
 			return err
 		}
 
-		oldNode := u.GetNodeFromProviderID(originalProviderID)
+		oldNode := u.GetNodeFromProviderID(originalProviderID.ID())
+		if oldNode == nil {
+			u.log.Info("Couldn't retrieve oldNode", "id", fmt.Sprintf("%q", originalProviderID))
```
ncdc: Do you need `fmt.Sprintf` here?

liztio: probably not, I was debugging whitespace issues.

pkg/upgrade/base.go
ncdc: Nevermind 😄

ncdc:
LGTM - would you please squash?

liztio:
@ncdc done

ncdc:
Thanks, doing a quick test locally before merging.

ncdc:
Ugh, can't integration test on Mac with the latest CAPD due to https://github.com/kubernetes-sigs/cluster-api-provider-docker/issues/69.

LGTM, will test with CAPA soon.

PR #67, opened 2019-08-07
Run tests with a pinned version of capdctl

### LINKED ISSUE #66: Make the tests work reliably
Right now there is no version control around what external tools are needed to run the tests. I'd like to make them reproducible so they don't break as work is done on #65

chuckha:

This fixes #66


chuckha:
I tested the make syntax and ran the MachineDeployment upgrade test and it passed

PR #68, opened 2019-09-05
[WIP] update to v1alpha2 related imports

amy:
Did a sweep to change v1alpha1 imports to v1alpha2.
(Still need to go through the tests.
Also need to sweep through the design docs to make sure there's nothing new missing for the upgrade path.)


pkg/upgrade/control_plane.go
```diff
@@ -13,19 +13,17 @@ import (
...
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	"github.com/vmware/cluster-api-upgrade-tool/pkg/internal/kubernetes"
 	v1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	clusterapiv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
-	"sigs.k8s.io/cluster-api/pkg/controller/noderefutil"
+	clusterapiv1alpha2 "sigs.k8s.io/cluster-api/api/v1alpha2"
```
chuckha: we've been naming these `clusterv1` as a convention. That way we don't have to change the references when we change the import path and it's consistent across cluster-api codebases

pkg/upgrade/control_plane.go
```diff
@@ -320,26 +316,17 @@ func (u *ControlPlaneUpgrader) updateMachines(machines *clusterapiv1alpha1.Machi
...
-	if err != nil {
-		return errors.WithStack(err)
-	}
-	if _, err := u.managementClusterAPIClient.Machines(m.Namespace).Patch(m.Name, types.MergePatchType, patch); err != nil {
-		return errors.WithStack(err)
-	}
+	updated := m.DeepCopy()
+
+	// TODO: double check this patch strategy
+	u.ctrlClient.Patch(context.TODO(), updated, ctrlclient.MergeFrom(original))
```
chuckha: check the error +wrap and return or return directly if applicable

pkg/upgrade/key_pair.go
```diff
@@ -14,8 +14,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/services/certificates"
```
chuckha: 🙏 been looking forward to this for a while

pkg/upgrade/machine_creator.go
```diff
@@ -77,9 +79,23 @@ func NewMachineCreator(options ...MachineCreatorOption) *MachineCreator {
 	return creator
 }
 
+func (n *MachineCreator) createMachine(machine *clusterapiv1alpha2.Machine) (*clusterapiv1alpha2.Machine, error) {
```
chuckha: there two methods can go away since we can fake the controller runtime client easily enough now

pkg/upgrade/machine_deployment.go
```diff
@@ -96,18 +95,9 @@ func (u *MachineDeploymentUpgrader) updateMachineDeployment(machineDeployment *c
...
-
-	// Create the patch
-	patchBytes, err := jsonpatch.CreateMergePatch(original, updated)
-	if err != nil {
-		return errors.Wrap(err, "error creating json patch")
-	}
+	updated := machineDeployment.DeepCopy()
 
-	_, err = u.managementClusterAPIClient.MachineDeployments(u.clusterNamespace).Patch(machineDeployment.Name, types.MergePatchType, patchBytes)
+	err := u.ctrlClient.Patch(context.TODO(), updated, ctrlclient.MergeFrom(original))
```
chuckha: instead of comparing against a deep copy this could compare against machineDeployment directly.

General convention we are using is that if a call returns just one error then it should be in the form of

```
if err := fn(); err != nil {
...
}
```

The benefit here is that we avoid the problem of accidentally not checking errors since this creates a new block. We don't run into:

```
x, err := fn()
// oops forgot to check error here but the compiler doesn't care because `err` is not an unused variable
y, err =  fn2()
if err != nil {}
```

chuckha:
EDIT: oops i just realized this was WIP. Sorry about that, please feel free to disregard my comments.

amy:
I messed up my git branches. Closing this PR and using this PR as the main one: https://github.com/vmware/cluster-api-upgrade-tool/pull/72

PR #80, opened 2019-10-03
Make the kubeconfig argument not required

### LINKED ISSUE #70: Assume an in cluster configuration if `--kubeconfig` is not specified
In cases where you are running the upgrade tool from the management cluster itself, it would be nice if the tool would use credentials from the service account. In client-go speak, using `InClusterConfig()` as opposed to specifying a file in the filesystem.

The workaround at the moment is creating a Configmap/Secret and mounting it into the upgrade pod. This works but is not ideal.

williamsandrew:
Make the `kubeconfig` argument not required which allows us to fall back to an in-cluster client configuration.

closes #70

Signed-off-by: Andrew Williams <andrewwi@vmware.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -40,8 +40,7 @@ func main() {
 	}
 
 	root.Flags().StringVar(&upgradeConfig.ManagementCluster.Kubeconfig, "kubeconfig",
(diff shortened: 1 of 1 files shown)


ncdc:
LGTM

PR #88, opened 2019-10-08
Only support proper kubeconfig secrets

### LINKED ISSUE #82: Remove old way of getting kubeconfig
See: https://github.com/vmware/cluster-api-upgrade-tool/pull/72/files#r329195623

* Remove `ca-secret` and `ca-field` flags.
* Update default for the KubeconfigSecretRef to reflect CAPI expectations

`ca-field` goes away since there is no more provider spec
`ca-secret` can go away even though we do store a ca in the secret we also store a kubeconfig. We only want to support using the kubeconfig

ncdc:
Remove support for specifying CA field and secret refs, and only support
the <cluster-name>-kubeconfig secret convention going forward.

Fixes #82


chuckha:
Is there any documentation that needs updating?

ncdc:
Good call on docs, probably. Checking

ncdc:
@chuckha updated

Issue #83, opened 2019-10-04
Use cluster-api patch helper


amy:
See: https://github.com/vmware/cluster-api-upgrade-tool/pull/72/files#r331558465

Currently patches are made with using the raw controller runtime. Replace those patches with the cluster-api helper.

PR #95, opened 2019-10-09
fix: remove old nodes from kubeadm configmap

### LINKED ISSUE #59: Update kubeadm configmap
When we remove a node we should remove it from the kubeadm config or risk etc attempting to contact a no longer responsive endpoint.

rudoi:
Fixes #59

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -340,7 +340,6 @@ modernc.org/golex v1.0.0/go.mod h1:b/QX9oBD/LhixY6NDh+IdGv17hgB+51fET1i2kPSmvk=
 modernc.org/mathutil v1.0.0/go.mod h1:wU0vUrJsVWBZ4P6e7xtFJEhFSNsfRLJ8H458uRjg03k=
 modernc.org/strutil v1.0.0/go.mod h1:lstksw84oURvj9y3tn8lGvRxyRC1S2+g5uuIzNfIOBs=
 modernc.org/xc v1.0.0/go.mod h1:mRNCo0bvLjGhHO9WsyuKVU4q0ceiDDDoEeWDJHrNx8I=
-sigs.k8s.io/cluster-api v0.2.5 h1:N3nz2U62tE1lm4szCrEDMPMg/8RMiLaYgHWZJqkSuLE=
 sigs.k8s.io/cluster-api v0.2.5/go.mod h1:Agc72Ra5LMOkQQ2v/Ywv1KUemaYAwvkQ+G59Ym5H8e4=
 sigs.k8s.io/cluster-api v0.2.6 h1:XG6qhUOirRiYs3BSGyijb+G/LmtK6uajVZyRieG2F3w=
 sigs.k8s.io/cluster-api v0.2.6/go.mod h1:Agc72Ra5LMOkQQ2v/Ywv1KUemaYAwvkQ+G59Ym5H8e4=
--- a/pkg/upgrade/control_plane.go
+++ b/pkg/upgrade/control_plane.go
@@ -42,9 +42,10 @@ import (
 )
 
 const (
-	etcdCACertFile = "/etc/kubernetes/pki/etcd/ca.crt"
-	etcdCertFile   = "/etc/kubernetes/pki/etcd/peer.crt"
-	etcdKeyFile    = "/etc/kubernetes/pki/etcd/peer.key"
+	etcdCACertFile       = "/etc/kubernetes/pki/etcd/ca.crt"
+	etcdCertFile         = "/etc/kubernetes/pki/etcd/peer.crt"
+	etcdKeyFile          = "/etc/kubernetes/pki/etcd/peer.key"
+	kubeadmConfigMapName = "kubeadm-config"
 
 	// annotationPrefix is the prefix for all annotations managed by this tool.
 	annotationPrefix = "upgrade.cluster-api.vmware.com/"
@@ -235,7 +236,7 @@ func (u *ControlPlaneUpgrader) updateKubeletConfigMapIfNeeded(version semver.Ver
 	log := u.log.WithValues("name", desiredKubeletConfigMapName)
 
 	log.Info("Checking for existence of kubelet configmap")
-	_, err := u.targetKubernetesClient.CoreV1().ConfigMaps("kube-system").Get(desiredKubeletConfigMapName, metav1.GetOptions{})
+	_, err := u.targetKubernetesClient.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(desiredKubeletConfigMapName, metav1.GetOptions{})
 	if err == nil {
 		log.Info("kubelet configmap already exists")
 		return nil
@@ -250,7 +251,7 @@ func (u *ControlPlaneUpgrader) updateKubeletConfigMapIfNeeded(version semver.Ver
 	previousMinorVersionKubeletConfigMapName := fmt.Sprintf("kubelet-config-%d.%d", version.Major, version.Minor-1)
 
 	log.Info("Retrieving kubelet configmap for previous minor version", "previous-name", previousMinorVersionKubeletConfigMapName)
-	cm, err := u.targetKubernetesClient.CoreV1().ConfigMaps("kube-system").Get(previousMinorVersionKubeletConfigMapName, metav1.GetOptions{})
+	cm, err := u.targetKubernetesClient.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(previousMinorVersionKubeletConfigMapName, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		return errors.Errorf("unable to find kubelet configmap %s", previousMinorVersionKubeletConfigMapName)
 	}
@@ -259,7 +260,7 @@ func (u *ControlPlaneUpgrader) updateKubeletConfigMapIfNeeded(version semver.Ver
 	cm.ResourceVersion = ""
 
 	log.Info("Creating kubelet configmap as a copy from the previous minor version")
-	_, err = u.targetKubernetesClient.CoreV1().ConfigMaps("kube-system").Create(cm)
+	_, err = u.targetKubernetesClient.CoreV1().ConfigMaps(metav1.NamespaceSystem).Create(cm)
 	if err != nil && !apierrors.IsAlreadyExists(err) {
 		return errors.Wrapf(err, "error creating configmap %s", desiredKubeletConfigMapName)
 	}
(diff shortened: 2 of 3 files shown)


pkg/upgrade/control_plane.go
```diff
@@ -419,6 +426,7 @@ func (u *ControlPlaneUpgrader) updateBootstrapConfig(name string, machine *clust
 		}
 	}
 	bootstrap.Spec.JoinConfiguration.NodeRegistration = nodeRegistration
+	bootstrap.Spec.JoinConfiguration.Discovery.BootstrapToken = &kubeadmv1beta1.BootstrapTokenDiscovery{}
```
amy: Is there an existing one to copy over?

pkg/upgrade/control_plane.go
```diff
@@ -277,6 +278,10 @@ func (u *ControlPlaneUpgrader) updateMachine(name string, machine clusterv1.Mach
 		return err
 	}
 
+	if err := u.applyAnnotation(newMachine); err != nil {
```
rudoi: I found most errors to show up during machine deletion / etcd member removal, so I put the annotation before those steps so that resuming would work.

rudoi: Though, you do still end up with an orphaned machine in either case.

pkg/upgrade/control_plane.go
chuckha: the problem with copying it over is that it's probably expired. Clearing it like this will generate a new one that won't be expired

pkg/upgrade/control_plane.go
```diff
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	bootstrapv1 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha2"
 	kubeadmv1beta1 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/kubeadm/v1beta1"
+	kubeadmv1beta2 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/kubeadm/v1beta2"
```
ncdc: Do you need v1beta2 specifically? It looks like the ClusterStatus type is the same in both versions?

pkg/upgrade/control_plane.go
```diff
@@ -700,3 +726,23 @@ func updateKubeadmKubernetesVersion(original *v1.ConfigMap, version string) (*v1
 
 	return cm, nil
 }
+
+func removeNodeFromClusterStatus(original *v1.ConfigMap, nodeName string) (*v1.ConfigMap, error) {
+	cm := original.DeepCopy()
+
+	clusterStatus := &kubeadmv1beta2.ClusterStatus{}
+	if err := yaml.Unmarshal([]byte(cm.Data["ClusterStatus"]), &clusterStatus); err != nil {
```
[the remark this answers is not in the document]
rudoi: Definitely, will fix. though we have other problems if this field is unset :D

pkg/upgrade/control_plane.go
```diff
@@ -700,3 +726,23 @@ func updateKubeadmKubernetesVersion(original *v1.ConfigMap, version string) (*v1
 
 	return cm, nil
 }
+
+func removeNodeFromClusterStatus(original *v1.ConfigMap, nodeName string) (*v1.ConfigMap, error) {
+	cm := original.DeepCopy()
+
+	clusterStatus := &kubeadmv1beta2.ClusterStatus{}
```
chuckha: we're using v1beta1 types from CABPK, is there any reason you imported v1beta2 instead?

pkg/upgrade/control_plane.go
```diff
@@ -681,6 +689,24 @@ func (u *ControlPlaneUpgrader) updateAndUploadKubeadmKubernetesVersion() error {
...
+	if err != nil {
+		return errors.Wrap(err, "error getting kubeadm configmap from target cluster")
+	}
+
+	updated, err := removeNodeFromClusterStatus(original, nodeName)
+	if err != nil {
+		return err
+	}
+
+	if _, err = u.targetKubernetesClient.CoreV1().ConfigMaps("kube-system").Update(updated); err != nil {
```
ncdc: Can we Patch instead of Update? That might help with the v1beta1 vs v1beta2 issue?

pkg/upgrade/control_plane.go
rudoi: I saw v1beta2 in my configmap so just went for that. will it work either way, you think?

rudoi: cc @chuckha

pkg/upgrade/control_plane.go
amy: oh cool

pkg/upgrade/control_plane.go
ncdc: I'd stick with v1beta1 if we can

pkg/upgrade/control_plane.go
amy: Example of patch: https://github.com/vmware/cluster-api-upgrade-tool/blob/master/pkg/upgrade/machine_deployment.go#L139

Though, there's a github issue to swap the raw controller runtime patches with cluster-api patch helper: https://github.com/vmware/cluster-api-upgrade-tool/issues/83

I'm cool with either.

rudoi: I think we'd have to patch if we wanted v1beta1, because the configmap itself has v1beta2 in the ClusterStatus field.

I'll try a patch. Time to remember how patching works with clientsets! :D

ncdc: Oh it's encoded yaml inside the configmap. Hmm. Patch won't help w/that. But unstructured.Unstructured will.

ncdc: Or create a new go struct that has the json tags you need on it and use that instead of the one in kubeadm.

chuckha: we're already using Update in the codebase for other kubeadm config map manipulation: https://github.com/vmware/cluster-api-upgrade-tool/blob/master/pkg/upgrade/control_plane.go#L684

maybe best to keep them the same for now?

ncdc: Yes, Update is fine to keep. Sorry for the noise.

pkg/upgrade/control_plane.go
detiber: :grimacing: should we track that as a separate issue?

pkg/upgrade/control_plane.go
rudoi: so do we want to do some unstructured goodies here? or what do we think would be most appropriate?

also, so I'm understanding, is this just so we can avoid importing the extra package?

ncdc: Mostly future-proofing so we don't have to fix this in the future if the api version changes but the structure remains the same

ncdc:
Woohoo CLA!

pkg/upgrade/control_plane.go
```diff
@@ -681,6 +688,24 @@ func (u *ControlPlaneUpgrader) updateAndUploadKubeadmKubernetesVersion() error {
 	return nil
 }
 
+func (u *ControlPlaneUpgrader) removeNodeFromKubeadmConfigMap(nodeName string) error {
+	original, err := u.targetKubernetesClient.CoreV1().ConfigMaps("kube-system").Get("kubeadm-config", metav1.GetOptions{})
```
neolit123: ideally `metav1.NamespaceSystem` should be used instead of the literal.
`kubeadm-config` should be constant too.

pkg/upgrade/control_plane.go
```diff
@@ -681,6 +688,24 @@ func (u *ControlPlaneUpgrader) updateAndUploadKubeadmKubernetesVersion() error {
...
+		return errors.Wrap(err, "error getting kubeadm configmap from target cluster")
+	}
+
+	updated, err := removeNodeFromClusterStatus(original, nodeName)
+	if err != nil {
+		return err
+	}
+
+	if _, err = u.targetKubernetesClient.CoreV1().ConfigMaps("kube-system").Update(updated); err != nil {
+		return errors.Wrap(err, "error updating kubeadm configmap")
```
neolit123: i would replace `configmap` with `ConfigMap` as per the sig-docs guide on API objects.

pkg/upgrade/control_plane.go
```diff
@@ -700,3 +725,32 @@ func updateKubeadmKubernetesVersion(original *v1.ConfigMap, version string) (*v1
 
 	return cm, nil
 }
+
+func removeNodeFromClusterStatus(original *v1.ConfigMap, nodeName string) (*v1.ConfigMap, error) {
+	cm := original.DeepCopy()
+
+	clusterStatus := &unstructured.Unstructured{}
+	if err := yaml.Unmarshal([]byte(cm.Data["ClusterStatus"]), &clusterStatus); err != nil {
+		return nil, errors.Wrap(err, "error decoding kubeadm configmap ClusterStatus")
```
neolit123: `error decoding the kubeadm ClusterStatus object`?

pkg/upgrade/control_plane.go
```diff
@@ -700,3 +725,32 @@ func updateKubeadmKubernetesVersion(original *v1.ConfigMap, version string) (*v1
...
+
+	delete(endpoints, nodeName)
+	err = unstructured.SetNestedMap(clusterStatus.UnstructuredContent(), endpoints, "apiEndpoints")
+	if err != nil {
+		return nil, err
+	}
+
+	updated, err := yaml.Marshal(clusterStatus)
+	if err != nil {
+		return nil, errors.Wrap(err, "error encoding kubeadm configmap ClusterConfiguration")
```
neolit123: `error encoding the kubeadm ClusterConfiguration object`?

pkg/upgrade/control_plane.go
rudoi: i globally swapped this in (in this file), FYI

go.sum
```diff
@@ -340,7 +340,6 @@ modernc.org/golex v1.0.0/go.mod h1:b/QX9oBD/LhixY6NDh+IdGv17hgB+51fET1i2kPSmvk=
 modernc.org/mathutil v1.0.0/go.mod h1:wU0vUrJsVWBZ4P6e7xtFJEhFSNsfRLJ8H458uRjg03k=
 modernc.org/strutil v1.0.0/go.mod h1:lstksw84oURvj9y3tn8lGvRxyRC1S2+g5uuIzNfIOBs=
 modernc.org/xc v1.0.0/go.mod h1:mRNCo0bvLjGhHO9WsyuKVU4q0ceiDDDoEeWDJHrNx8I=
-sigs.k8s.io/cluster-api v0.2.5 h1:N3nz2U62tE1lm4szCrEDMPMg/8RMiLaYgHWZJqkSuLE=
 sigs.k8s.io/cluster-api v0.2.5/go.mod h1:Agc72Ra5LMOkQQ2v/Ywv1KUemaYAwvkQ+G59Ym5H8e4=
```
vincepri: This line should also get removed with `make module`

ncdc: No - cabpk still uses v0.2.5 (we chatted on Slack about this)

pkg/upgrade/control_plane.go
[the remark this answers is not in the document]
ncdc: @rudoi feel like doing the nil check just to be safe?

go.sum
vincepri: ah gotcha, sounds good

pkg/upgrade/control_plane.go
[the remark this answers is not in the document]
rudoi: yep, no problem. missed this one.

go.sum
rudoi: ran `make modules`, no diff FYI

rudoi:
@vincepri did I miss any? grabbed the nil check.

vincepri:
that was the one!

ncdc:
@rudoi could you please squash?

rudoi:
> @rudoi could you please squash?

done

PR #101, opened 2019-10-17
Update CAPI/CABPK/controller-runtime deps

akutz:
This patch updates the following dependencies:

* CAPI v0.2.5
* CABPK v0.1.3
* controller-runtime v0.3.0

Related to https://github.com/kubernetes-sigs/cluster-api/issues/1598


go.mod
```diff
@@ -11,20 +11,20 @@ require (
...
-	sigs.k8s.io/cluster-api v0.2.1
-	sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm v0.1.0
-	sigs.k8s.io/controller-runtime v0.2.2
+	sigs.k8s.io/cluster-api v0.2.5
+	sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm v0.1.3
+	sigs.k8s.io/controller-runtime v0.3.0
 	sigs.k8s.io/yaml v1.1.0
 )
 
 replace (
```
ncdc: We shouldn't need `replace` any more

ncdc:
Also please run `make modules` after removing the `replace` content 😄

akutz:
> Also please run `make modules` after removing the `replace` content 😄

Hi @ncdc,

I will. I did. That's the strange thing. I'm used to the pattern. I was also surprised by that error.

ncdc:
Yeah, that would have been my next recommendation if there were still issues. Thanks for doing this!

PR #103, opened 2019-10-17
Remove MachineDeployment support

ncdc:
Fixes #102

PR #104, opened 2019-10-17
Support a --config file

### LINKED ISSUE #38: accept a config file
The config struct is already built out. As an additional option, we should allow users to pass a config file to the command line tool instead of specifying a bunch of args. It can be a JSON config file for now if that makes this issue easier. But I think YAML shouldn't be a huge pain to implement.

ncdc:
Only the last commit is relevant - will rebase after #103 is merged

Fixes #38

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,10 +8,12 @@ import (
 	"os"
 
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/vmware/cluster-api-upgrade-tool/pkg/logging"
 	"github.com/vmware/cluster-api-upgrade-tool/pkg/upgrade"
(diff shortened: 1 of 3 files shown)


ncdc:
Rebased

main.go
```diff
@@ -22,12 +24,30 @@ func newLogger() logr.Logger {
...
+				}
+				defer reader.Close()
+
+				const bufferSize = 512
+				decoder := yaml.NewYAMLOrJSONDecoder(reader, bufferSize)
+				if err := decoder.Decode(&upgradeConfig); err != nil {
+					return errors.Wrap(err, "error parsing config file")
+				}
+
+				// TODO either merge in other flag values after decoding, or raise an error if the user specifies
```
chuckha: let's turn this into an issue so we can discuss approaches

chuckha: (i will do this)

ncdc:
@chuckha I updated the readme!

PR #105, opened 2019-10-17
Support infra and bootstrap patches

### LINKED ISSUE #99: Add support for adjusting fields in referenced infra and bootstrap objects for control plane machines
We added code originally to be able to modify the ID of a Machine's image (AMI/OVA/etc), in case the user wanted to specify the ID rather than letting the provider choose it based on the Kubernetes version.

With the conversion to only support CAPI v1alpha2, the image field is no longer part of the providerSpec blob in the Machine spec. If one exists, it's now in the infrastructure provider's machine resource (e.g. `AWSMachine`, for CAPA). The code we have doesn't work any more, since it still acts against the Machine for modifying the image.

Do we still need to be able to specify the Machine's image ID when doing an upgrade? If so, we'll need to adjust the code to mutate the object referenced by Machine.Spec.InfrastructureRef.

ncdc:
Fixes #99

Only the last commit is relevant - will rebase after #103 and #104 merge


pkg/upgrade/config.go
```diff
@@ -14,6 +14,7 @@ type Config struct {
 	TargetCluster     TargetClusterConfig     `json:"targetCluster"`
 	KubernetesVersion string                  `json:"kubernetesVersion"`
 	UpgradeID         string                  `json:"upgradeID"`
+	Patches           PatchConfig             `json:"patches"`
```
chuckha: should this be a slice of patch configs? or is this literally supposed to be `[{//patchop}, {//patchop}, ...]`?

chuckha: ah never mind, the patch object is actually a list of operations.

pkg/upgrade/control_plane.go
```diff
@@ -733,21 +786,29 @@ func (u *ControlPlaneUpgrader) updateInfrastructureReference(replacementKey ctrl
...
+	if u.infrastructurePatch != nil {
+		toCreate, err = patchRuntimeObject(infra, u.infrastructurePatch)
+		if err != nil {
+			return errors.Wrap(err, "error patching infrastructure resource")
+		}
+	}
 
 	// create the replacement infrastructure object
-	err = u.managementClusterClient.Create(context.TODO(), infraRef)
+	err = u.managementClusterClient.Create(context.TODO(), toCreate)
```
chuckha: weird. I got a strange stack trace here:

```
INFO[0000] Updating infrastructure reference             api-version=infrastructure.cluster.x-k8s.io/v1alpha2 kind=DockerMachine machine=default/controlplane-0 name=controlplane-0 upgrade-id=1
Error: the server could not find the requested resource
the server could not find the requested resource
github.com/vmware/cluster-api-upgrade-tool/pkg/upgrade.(*ControlPlaneUpgrader).updateInfrastructureReference
	/home/ubuntu/cluster-api-upgrade-tool/pkg/upgrade/control_plane.go:813
github.com/vmware/cluster-api-upgrade-tool/pkg/upgrade.(*ControlPlaneUpgrader).updateMachines
	/home/ubuntu/cluster-api-upgrade-tool/pkg/upgrade/control_plane.go:571
github.com/vmware/cluster-api-upgrade-tool/pkg/upgrade.(*ControlPlaneUpgrader).Upgrade
	/home/ubuntu/cluster-api-upgrade-tool/pkg/upgrade/control_plane.go:216
main.main.func1
	/home/ubuntu/cluster-api-upgrade-tool/main.go:56
github.com/spf13/cobra.(*Command).execute
	/home/ubuntu/go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:762
github.com/spf13/cobra.(*Command).ExecuteC
	/home/ubuntu/go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:852
github.com/spf13/cobra.(*Command).Execute
	/home/ubuntu/go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:800
main.main
	/home/ubuntu/cluster-api-upgrade-tool/main.go:118
runtime.main
	/home/ubuntu/.gimme/versions/go1.12.12.linux.amd64/src/runtime/proc.go:200
runtime.goexit
	/home/ubuntu/.gimme/versions/go1.12.12.linux.amd64/src/runtime/asm_amd64.s:1337
```

I would expect "could not create" but not "could not find" 🤔

ncdc: I think this means it couldn't find the `dockermachine` CRD type. Maybe a bug in this code somewhere?

ncdc: Oh I wonder if we lost the typemeta

chuckha: definitely lost typemeta.

Issue #106, opened 2019-10-18
Update go.mod to latest CAPI, CABPK


ncdc:
CAPI v0.2.6
CABPK v0.1.4

PR #107, opened 2019-10-18
Update go.mod to latest CAPI, CABPK

noamran:
Fixes: #106 

CAPI v0.2.6
CABPK v0.1.4


ncdc:
Thanks! If you run `make modules` does it remove the old versions from go.sum?

ncdc:
Yeah, need to run `make modules` to pass CI

PR #114, opened 2019-10-21
Require upgrade ID

### LINKED ISSUE #108: upgrade-id is marked as optional but is required
We can either default it to a value or mark it as required. I have no preference in the direction this goes so long as the tool's help is consistent with its behavior.

Right now we end up with a stack trace when the upgrade-id (marked optional) is omitted.

ncdc:
Fixes #108

### DIFF
--- a/main.go
+++ b/main.go
@@ -98,7 +98,7 @@ func main() {
 		&upgradeConfig.UpgradeID,
 		"upgrade-id",
 		"",
-		"Unique identifier used to resume a partial upgrade (optional)",
+		"Unique identifier used to resume a partial upgrade (required)",
 	)
 
 	root.Flags().StringVar(
--- a/pkg/upgrade/control_plane.go
+++ b/pkg/upgrade/control_plane.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -82,8 +83,13 @@ func NewControlPlaneUpgrader(log logr.Logger, config Config) (*ControlPlaneUpgra
 	if config.KubernetesVersion == "" {
 		return nil, errors.New("kubernetes version is required")
 	}
-	if !upgradeIDInputRegex.MatchString(config.UpgradeID) {
-		return nil, errors.New("upgrade ID must be a timestamp containing only digits")
+	if config.UpgradeID == "" {
+		return nil, errors.New("upgrade ID is required")
+	}
+	// Kubernetes resource names must be DNS1123 subdomains. Because the upgrade ID becomes part of the name for new
+	// machines, infra machines, and KubeadmConfigs, we use the same validation here.
+	if errs := validation.IsDNS1123Subdomain(config.UpgradeID); len(errs) > 0 {
+		return nil, errors.New("upgrade ID: " + strings.Join(errs, ", "))
 	}
 
 	var userVersion, desiredVersion semver.Version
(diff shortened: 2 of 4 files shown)


pkg/upgrade/control_plane.go
```diff
@@ -82,8 +82,11 @@ func NewControlPlaneUpgrader(log logr.Logger, config Config) (*ControlPlaneUpgra
 	if config.KubernetesVersion == "" {
 		return nil, errors.New("kubernetes version is required")
 	}
-	if !upgradeIDInputRegex.MatchString(config.UpgradeID) {
-		return nil, errors.New("upgrade ID must be a timestamp containing only digits")
+	if config.UpgradeID == "" {
+		return nil, errors.New("upgrade ID is required")
+	}
+	if errs := validation.IsDNS1123Subdomain(config.UpgradeID); len(errs) > 0 {
```
chuckha: Can you add a comment around why this is dns 1123 subdomain? I can't recall RFC 1123 off the top of my head. I assume this is used because it's conveniently already written and tested?

ncdc: Updated w/a comment

ncdc:
I need to fix the regex - I made changes to open up the upgrade ID to more than just numbers, but missed it when checking for the machine name suffix

ncdc:
@chuckha I just fixed `upgradeIDNameSuffixRegex` to match the loosened restriction on upgrade id format.

main.go
```diff
@@ -98,7 +98,7 @@ func main() {
 		&upgradeConfig.UpgradeID,
 		"upgrade-id",
 		"",
-		"Unique identifier used to resume a partial upgrade (optional)",
+		"Unique identifier used to resume a partial upgrade (required)",
```
vincepri: You should also be able to use https://godoc.org/github.com/spf13/cobra#MarkFlagRequired

ncdc: We had that originally, but we can't use it because the flags themselves are not required when using `--config`

vincepri: Got it, make sense!

pkg/upgrade/control_plane.go
```diff
@@ -594,6 +596,9 @@ func upgradeSuffix(upgradeID string) string {
 	return ".upgrade." + upgradeID
```
chuckha: Can this break `validation.IsDNS1123Subdomain`? For example, what happens if my upgrade id is `.123` which would make this `.upgrade..123`. Is that valid?

ncdc: Good question! `.123` won't pass that validation, and therefore is not a valid upgrade id
```
go run ./main.go --cluster-namespace cns --cluster-name cn --kubernetes-version v1.15.3 --upgrade-id .123
Error: upgrade ID: a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
upgrade ID: a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
github.com/vmware/cluster-api-upgrade-tool/pkg/upgrade.NewControlPlaneUpgrader
        /Users/andy/code/capi-dev/cluster-api-upgrade-tool/pkg/upgrade/control_plane.go:92
main.main.func1
        /Users/andy/code/capi-dev/cluster-api-upgrade-tool/main.go:51
```

chuckha: oh, wonderful

PR #115, opened 2019-10-21
Make CLI flags override config file values

### LINKED ISSUE #111: config file vs command line flags
with the addition of #104 we will be accepting config files in addition to command line flags.

It is a bit ambiguous how these two user supplied fields will interact with each other.

I suggest having the command line flags override the config file. The reason for this is that sometimes it's nice to use the same config with stuff I don't want to manually type in but override just a single field, such as version. Whatever the implementation, I'd like to avoid defining a complex precedence order.

ncdc:
Fixes #111

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/blang/semver v3.5.0+incompatible
 	github.com/evanphx/json-patch v4.5.0+incompatible
 	github.com/go-logr/logr v0.1.0
+	github.com/imdario/mergo v0.3.8
 	github.com/pkg/errors v0.8.1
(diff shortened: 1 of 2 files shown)


chuckha:
Did you test this out?

ncdc:
I did, although I'm wondering if maybe we should refactor `main.go` a bit to make this testable

chuckha:
we could...i don't know if this logic is complex enough to warrant it though. i'd be more interested in putting that time to re building the e2e tests.

ncdc:
SGTM! (I was mainly referring to the config merging part)

PR #118, opened 2019-10-23
Retry machine deletion up to 5 minutes

### LINKED ISSUE #96: apiserver pod for removed etcd member remains, potentially causing subsequent machine deletion to fail
So the current flow per controlplane machine is, simplified:
1. remove etcd member
2. delete Machine

Technically, between those two steps, there is an instance still being targeted by the ELB that does not have etcd behind it. API Server requests made that would be routed to this apiserver instance will time out. Sometimes the delete machine call (step 2) fails because of this.

Possible things:
1. delete the apiserver pod on the old node before removing the etcd member
2. wait for ELB to declare instance unhealthy (how?)

Additionally:
1. per discussion with @chuckha, there may be some CAPA fixes involved here: should CAPA ensure ELB targets are removed when controlplane Machines are deleted?

cc @amy

ncdc:
Retry machine deletion up to 5 minutes, in case the cluster's load
balancer routes the request to the machine being deleted.

Fixes #96

### DIFF
--- a/pkg/upgrade/control_plane.go
+++ b/pkg/upgrade/control_plane.go
@@ -468,10 +468,21 @@ func (u *ControlPlaneUpgrader) updateMachine(replacementKey ctrlclient.ObjectKey
 		}
 	}
 
+	const (
+		deleteMachineInterval = 10 * time.Second
+		deleteMachineTimeout  = 5 * time.Minute
+	)
 	u.log.Info("Deleting existing machine", "namespace", machine.Namespace, "name", machine.Name)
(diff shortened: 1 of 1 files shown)


ncdc:
@rudoi PTAL - I think this is about all we can do for now

detiber:
travis doesn't appear happy with the state of go modules

ncdc:
Fixed. I wonder if there's a mismatch between code in mainline vs test/integration (build tag)

rudoi:
this is totally fine by me - figured this is where it would land :)

PR #119, opened 2019-10-24
Fix main go.mod for client-go

ncdc:
Move the integration test into its own separate go module so it can
import capd without messing up the main module's client-go version.


Makefile
```diff
@@ -56,20 +56,13 @@ lint-full: $(GOLANGCI_LINT) ## Run slower linters to detect possible issues
...
-endif
-
 .PHONY: test
 test:  ## Run unit tests
 	go test ./...
 
 .PHONY: integration-test
 integration-test: ## Run integration tests
-	go test -tags integration -count=1 -v -timeout=20m $(TEST_ARGS) ./test/integration
+	cd test/integration; go test -count=1 -ginkgo.v -v -timeout=20m $(TEST_ARGS)
```
chuckha: `-ginkgo.v`?

chuckha: ah there it is. That was super hard to google.

PR #122, opened 2019-10-25
Poll for machine deletion and kubeadm configmap updates

ncdc:
Fixes #120

PR #125, opened 2019-10-28
Update kubeadm ConfigMap apiEndpoints

ncdc:
Update kubeadm ConfigMap ClusterStatus.APIEndpoints at the beginning and
end of the upgrade process.

Fixes #122


ncdc:
I need to test this before we merge it

ncdc:
Tested locally. Found a bug. Fixed it. Would appreciate another manual test - @chuckha do you have time today or tomorrow?

pkg/upgrade/control_plane.go
```diff
@@ -1043,6 +1059,8 @@ func (u *ControlPlaneUpgrader) removeNodeFromKubeadmConfigMap(nodeName string) e
 	return nil
 }
 
+// updateKubeadmKubernetesVersion updates the Kubernetes version stored in the kubeadm configmap. This is
+// // required so that new Machines joining the cluster use the correct Kubernetes version as part of the upgrade.
```
vincepri:
```suggestion
// required so that new Machines joining the cluster use the correct Kubernetes version as part of the upgrade.
```

pkg/upgrade/control_plane.go
```diff
@@ -175,6 +175,20 @@ func (u *ControlPlaneUpgrader) Upgrade() error {
...
+	nodeList, err := u.targetKubernetesClient.CoreV1().Nodes().List(metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "error listing workload cluster nodes")
+	}
+	u.log.Info("Reconciling kubeadm ConfigMap's ClusterStatus.APIEndpoints")
+	if err := u.updateKubeadmConfigMap(func(in *v1.ConfigMap) (*v1.ConfigMap, error) {
+		return reconcileKubeadmConfigMapClusterStatusAPIEndpoints(in, nodeList, machines)
+	}); err != nil {
+		return errors.Wrap(err, "error reconciling kubeadm ConfigMap")
+	}
```
vincepri: If we move all of this logic into its own method, we could just call it twice instead of replicating it in two places

ncdc: Good idea, will do

ncdc:
@chuckha I can do the custom type. The code's been bugging me too!

chuckha:
Worked well on a multi-control-plane cluster!

PR #126, opened 2019-10-29
Check etcd cluster health before each machine

### LINKED ISSUE #123: lost master during upgrade
One of my master nodes was unable to bring up etcd fast enough:

Here are the relevant lines in /var/log/cloud-init-output.log
```
...
[etcd] Announced new etcd member joining to the existing etcd cluster
[etcd] Wrote Static Pod manifest for a local etcd member to "/etc/kubernetes/manifests/etcd.yaml"
[etcd] Waiting for the new etcd member to join the cluster. This can take up to 40s
[kubelet-check] Initial timeout of 40s passed.
error execution phase control-plane-join/etcd: error creating local etcd static pod manifest file: timeout waiting for etcd cluster to be available
```

This kinda sucks because it's not like we can re-run kubeadm init :/ I don't know if there's anything we can really do in this situation, at least not in this repo. Feel free to close if this is too inactionable.

### LINKED ISSUE #124: Wait for etcd cluster health before proceeding to the next machine
`kubeadm` waits for a new etcd pod to be "ready" (as defined by Kubernetes, which means the pod's container is running and probes are passing), but etcd could still be catching up the new member and settling. Before proceeding with each new machine, we should check etcd cluster health, waiting/retrying up to _n_ minutes before giving up.

ncdc:
Fixes #124 
Fixes #123

### DIFF
--- a/pkg/upgrade/control_plane.go
+++ b/pkg/upgrade/control_plane.go
@@ -192,7 +192,7 @@ func (u *ControlPlaneUpgrader) Upgrade() error {
 	}
 
 	u.log.Info("Checking etcd health")
-	if err := u.etcdClusterHealthCheck(time.Minute * 1); err != nil {
+	if err := u.etcdClusterHealthCheck(15 * time.Second); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/upgrade/control_plane.go
```diff
@@ -616,6 +616,18 @@ func (u *ControlPlaneUpgrader) updateMachines(machines []*clusterv1.Machine) err
 
 		// TODO skip if the bootstrap ref is not a KubeadmConfig
 
+		log.Info("Checking etcd health")
+		err := wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
+			if err := u.etcdClusterHealthCheck(15 * time.Second); err != nil {
```
detiber: In `etcdClusterHealthCheck`, could we use the `--cluster` flag instead of `--endpoints` now, since we are no longer supporting v1.13?

Otherwise, lgtm

ncdc: I think I'd feel more comfortable leaving the `--endpoints` bit for now, since we know it works.

detiber: Sounds good, should I create an issue to track following up or are we okay with leaving it as is, since v1alpha3 should obviate the need for this tool?

ncdc: I'm ok leaving as-is w/v1a3 looming

PR #134, opened 2019-11-13
clear replacement machine status to ensure no matches to previous machines

andrewsykim:
CAPV doesn't actually provision a new machine based on the copied replacement machine because it assumes one already exists by checking the machine status. Machine status should be cleared anyways given everything in status is specific to the previously copied machine.

Signed-off-by: Andrew Sy Kim <kiman@vmware.com>

### DIFF
--- a/pkg/upgrade/control_plane.go
+++ b/pkg/upgrade/control_plane.go
@@ -458,6 +458,9 @@ func (u *ControlPlaneUpgrader) updateMachine(replacementKey ctrlclient.ObjectKey
 		desiredVersion := u.desiredVersion.String()
 		replacementMachine.Spec.Version = &desiredVersion
 
+		// clear status to ensure the provider can't match the new machine to any existing machine
+		replacementMachine.Status = clusterv1.MachineStatus{}
+
 		log.Info("Creating new machine")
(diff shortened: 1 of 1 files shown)


pkg/upgrade/control_plane.go
```diff
@@ -458,6 +458,9 @@ func (u *ControlPlaneUpgrader) updateMachine(replacementKey ctrlclient.ObjectKey
 		desiredVersion := u.desiredVersion.String()
 		replacementMachine.Spec.Version = &desiredVersion
 
+		// clear status to ensure the provider can't match the new machine to any existing machine
```
ncdc: @andrewsykim I don't believe this is necessary. The apiserver will throw away anything in the .status portion of the object as part of the create flow. Are you seeing otherwise?

andrewsykim: I am seeing otherwise but it may be CAPV doing that and not the apiserver -- will verify in a bit

andrewsykim: yeah so turns out it was CAPV re patching the status there so we don't need this, thanks!

/close

PR #142, opened 2019-11-19
Fix kubeadm api endpoints logic

### LINKED ISSUE #141: kubeadm configmap cluster status api endpoints sync code is buggy
The code currently reconciles between Nodes in the workload cluster and Machines. If it finds a Node that is being deleted or doesn't have a matching Machine, it removes the entry from the map. This is fine for Nodes that still exist... but if both the Node and its associated Machine have been deleted, this logic makes it impossible to remove an entry for this Node from the ConfigMap if it's still there for whatever reason.

A better alternative could be to remove all entries from the ConfigMap where this is not a Node with a matching name. The entry in the map is the Node's name, so this is a 1:1 mapping. This is predicated on node bootstrapping doing the following:

1. `kubeadm join` runs
    1. Node is registered in etcd
    1. ConfigMap is updated with new Node

If this is the sequence, we are good. If, however, the ConfigMap is updated first, there is a chance for a race condition where the entry exists in the ConfigMap, there is no Node in etcd yet, and the upgrade tools runs and reconciles the data, removing the entry from the map because it doesn't see a corresponding node.

cc @fabriziopandini - let's chat about this next week!

ncdc:
Instead of matching nodes and machines, which may cause us to skip over
invalid entries in the map, match between map entries and nodes,
deleting any entry that doesn't have a matching node.

Fixes #141

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 	k8s.io/api v0.0.0-20190918195907-bd6ac527cfd2
 	k8s.io/apimachinery v0.0.0-20190817020851-f2f3a405f61d
 	k8s.io/client-go v0.0.0-20190918200256-06eb1244587a
-	k8s.io/utils v0.0.0-20190809000727-6c36bc71fc4a
 	sigs.k8s.io/cluster-api v0.2.6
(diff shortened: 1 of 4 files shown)


ncdc:
@williamsandrew can you try this out? It worked for me in testing this morning.

PR #145, opened 2019-12-04
Add kubeadm configmap patching support

### LINKED ISSUE #144: The --bootstrap-patches option only changes the KubeadmConfig API but not Configmap in workload cluster
When supplying a set of JSON patches via the `--bootstrap-patches` option the KubeadmConfig API object in the management cluster is updated successfully. However, the Kubeadm configmap in the workload cluster is not changed which means said patches are effectively ignored. The only change that happens to the Kubeadm configmap is the `version` field which comes from this code: https://github.com/vmware/cluster-api-upgrade-tool/blob/0288bc748c450b487caceae130a9822df5d47a46/pkg/upgrade/control_plane.go#L204-L206.

To replicate, run the upgrade tool with a `--bootstrap-patches` option similar to this:

```
--bootstrap-patches=[{"op":"replace","path":"/spec/clusterConfiguration/dns/imageRepository","value":"example.com"},{"op":"replace","path":"/spec/clusterConfiguration/dns/imageTag","value":"some-tag"}]
```

Now review the KubeadmConfig API object for the upgraded machine as well as the Kubeadm Configmap in the workload cluster. You'll notice that only the former changed the DNS options.

ncdc:
Fixes #144

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,16 +85,17 @@ Usage:
   cluster-api-upgrade-tool [flags]
 
 Flags:
-      --bootstrap-patches string        JSON patch expression of patches to apply to the machine's bootstrap resource (optional)
-      --cluster-name string             The name of target cluster (required)
-      --cluster-namespace string        The namespace of target cluster (required)
-      --config string                   Path to a config file in yaml or json format
-  -h, --help                            help for cluster-api-upgrade-tool
-      --infrastructure-patches string   JSON patch expression of patches to apply to the machine's infrastructure resource (optional)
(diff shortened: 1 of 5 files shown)


ncdc:
@williamsandrew I picked your commit

williamsandrew:
Do you want to me to adjust the commit message and re-pick since it has `WIP` in there?

ncdc:
It's really up to you

williamsandrew:
LGTM

PR #151, opened 2020-02-15
Reconcile bad replacement

### LINKED ISSUE #149: Upgrade completes successfully but fewer masters than expected
Our team is seeing an occasional issue where an upgrade completes successfully according to the upgrade tool but when you look at the `kubectl get node` output there will be fewer master nodes than expected. This appears to be related to `kubeadm` failing at the Etcd step. See below:

```
[etcd] Announced new etcd member joining to the existing etcd cluster
[etcd] Creating static Pod manifest for "etcd"
[etcd] Waiting for the new etcd member to join the cluster. This can take up to 40s
{"level":"warn","ts":"2020-02-06T18:06:40.249Z","caller":"clientv3/retry_interceptor.go:61","msg":"retrying of unary invoker failed","target":"passthrough:///https://10.0.3.21:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = context deadline exceeded"}
[kubelet-check] Initial timeout of 40s passed.
error execution phase control-plane-join/etcd: error creating local etcd static pod manifest file: timeout waiting for etcd cluster to be available
To see the stack trace of this error execute with --v=5 or higher
Feb 06 18:09:35 cloud-init[4206]: util.py[WARNING]: Failed running /var/lib/cloud/instance/scripts/runcmd [1]
Feb 06 18:09:35 cloud-init[4206]: cc_scripts_user.py[WARNING]: Failed to run module scripts-user (scripts in /var/lib/cloud/instance/scripts)
Feb 06 18:09:35 cloud-init[4206]: util.py[WARNING]: Running module scripts-user (<module 'cloudinit.config.cc_scripts_user' from '/usr/lib/python2.7/site-packages/cloudinit/config/cc_scripts_user.pyc'>) failed
Cloud-init v. 19.3-2.amzn2 finished at Thu, 06 Feb 2020 18:09:35 +0000. Datasource DataSourceEc2.  Up 227.89 seconds
```

However, the upgrade tool continues onto the next node despite this failure. If you look at the attached log output, it *does* appear that the new Etcd member joined the cluster despite the Deadline above.

Here is the out of `kubectl get nodes` on this cluster. The "ip-10-0-3-21.ec2.internal" is supposed to be the 3rd master node.

```
$ kubectl get nodes
NAME                         STATUS   ROLES    AGE   VERSION
ip-10-0-1-18.ec2.internal    Ready    <none>   21h   v1.16.4+vmware.1
ip-10-0-1-207.ec2.internal   Ready    master   21h   v1.16.4+vmware.1
ip-10-0-3-21.ec2.internal    Ready    <none>   21h   v1.16.4+vmware.1
ip-10-0-5-23.ec2.internal    Ready    master   21h   v1.16.4+vmware.1
```

[logs.txt](https://github.com/vmware/cluster-api-upgrade-tool/files/4171540/logs.txt)

wfernandes:
A few things I would like feedback on:

- Verify logs are sufficient and have appropriate tags
- Verify context timeouts and other wait timeouts are sufficient and will not prolong an upgrade process
- I moved `reconcileMachineAnnotations` to run before reconciling replacement machines as well. After a few test runs, it doesn't look like it has any negative side-effects.
- In the `removeMachine` method, I changed the order of removal. Not sure if this matters. From:
  1. delete the EtcdMember
  1. remove existing Machine
  1. remove machine from kubeadm configmap

- to:
  1. remove existing Machine
  1. delete the EtcdMember
  1. remove machine from kubeadm configmap

Fixes #149

### DIFF
--- a/pkg/upgrade/control_plane.go
+++ b/pkg/upgrade/control_plane.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -47,6 +48,7 @@ const (
 	etcdCertFile         = "/etc/kubernetes/pki/etcd/peer.crt"
 	etcdKeyFile          = "/etc/kubernetes/pki/etcd/peer.key"
 	kubeadmConfigMapName = "kubeadm-config"
+	LabelNodeRoleMaster  = "node-role.kubernetes.io/master"
 
 	// annotationPrefix is the prefix for all annotations managed by this tool.
 	annotationPrefix = "upgrade.cluster-api.vmware.com/"
@@ -159,6 +161,8 @@ func NewControlPlaneUpgrader(log logr.Logger, config Config) (*ControlPlaneUpgra
 		}
 	}
 
+	log.Info(fmt.Sprintf("Running upgrade with following configuration: %+v", config))
+
 	return &ControlPlaneUpgrader{
 		log:                     log,
 		desiredVersion:          parsedVersion,
@@ -185,6 +189,7 @@ func (u *ControlPlaneUpgrader) Upgrade() error {
 	if len(machines) == 0 {
 		return errors.New("Found 0 control plane machines")
 	}
+	u.log.Info("Found control plane machines", "count", len(machines))
 
 	// Begin the upgrade by reconciling the kubeadm ConfigMap's ClusterStatus.APIEndpoints, just in case the data
 	// is out of sync.
@@ -252,6 +257,25 @@ func (u *ControlPlaneUpgrader) Upgrade() error {
 		return err
 	}
 
+	// reconcile machine annotation in the beginning for all the machines as
+	// part of this upgrade process.
+	for _, m := range machines {
+		if err := u.reconcileMachineAnnotations(m); err != nil {
+			u.log.Error(err, "Unable to reconcile machine annotations",
+				"machine",
+				fmt.Sprintf("%s/%s", m.Namespace, m.Name),
+			)
+			return err
+		}
+	}
+
+	// handle all replacement machines first in case they are in an unhealthy
+	// state.
+	u.log.Info("Reconciling replacement machines")
+	if err := u.reconcileReplacements(machines); err != nil {
+		return err
+	}
+
 	u.log.Info("Updating machines")
 	if err := u.updateMachines(machines); err != nil {
 		return err
@@ -267,7 +291,14 @@ func (u *ControlPlaneUpgrader) Upgrade() error {
 			Name:      replacementName,
 		}
 
-		if err := u.managementClusterClient.Get(context.TODO(), key, &replacement); err != nil {
+		if err := u.managementClusterClient.Get(context.TODO(), key, &replacement); apierrors.IsNotFound(err) {
+			// if there is a bad replacement machine in the beginning of the
+			// upgrade it is eventually deleted but the reference of the
+			// machine object is still in the slice of machines so just log
+			// this error and continue.
+			u.log.Info("skipping annotation removal from machine as it no longer exists", "machine", key.String())
+			continue
+		} else if err != nil {
 			return errors.Wrapf(err, "error getting machine %s", key.String())
 		}
 
@@ -449,21 +480,6 @@ func (u *ControlPlaneUpgrader) updateMachine(replacementKey ctrlclient.ObjectKey
 		"replacement", replacementKey.String(),
 	)
 
-	originalProviderID, err := noderefutil.NewProviderID(*machine.Spec.ProviderID)
-	if err != nil {
-		return err
-	}
-	log.Info("Determined provider id for machine", "provider-id", originalProviderID)
-
-	oldNode := u.GetNodeFromProviderID(originalProviderID.ID())
-	if oldNode == nil {
-		u.log.Info("Couldn't retrieve oldNode", "id", originalProviderID.String())
-		return fmt.Errorf("unknown previous node %q", originalProviderID.String())
-	}
-
-	oldNodeName := oldNode.Name
-	log.Info("Got node name for machine", "node", oldNode.Name)
-
 	log.Info("Checking if we need to create a new machine")
 	replacementRef := v1.ObjectReference{
 		APIVersion: clusterv1.GroupVersion.String(),
@@ -509,20 +525,7 @@ func (u *ControlPlaneUpgrader) updateMachine(replacementKey ctrlclient.ObjectKey
 		}
 	}
 
-	ctx := context.TODO()
-	ctx, cancel := context.WithTimeout(ctx, u.machineTimeout)
-	defer cancel()
-
-	log.Info("Waiting for new machine", "timeout", u.machineTimeout)
-	newProviderID, err := u.waitForProviderID(ctx, u.clusterNamespace, replacementKey.Name)
-	if err != nil {
-		return err
-	}
-	node, err := u.waitForMatchingNode(ctx, newProviderID)
-	if err != nil {
-		return err
-	}
-	if err := u.waitForNodeReady(ctx, node); err != nil {
+	if err := u.isHealthy(replacementMachine); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


wfernandes:
/do-not-merge 
Need to make a change regarding timeouts.

pkg/upgrade/control_plane.go
```diff
@@ -268,7 +289,8 @@ func (u *ControlPlaneUpgrader) Upgrade() error {
 		}
 
 		if err := u.managementClusterClient.Get(context.TODO(), key, &replacement); err != nil {
-			return errors.Wrapf(err, "error getting machine %s", key.String())
+			u.log.Error(err, "error getting machine", "machine", key.String())
```
wfernandes: Instead of returning an error here, we just log it and continue. 
If there is a bad replacement machine in the beginning, it is then deleted but the reference of the machine object is still in the slice of `machines`. 
The original behavior would return an error when trying to remove upgrade annotations but we just want it to log it and move on.

pkg/upgrade/control_plane.go
```diff
@@ -673,18 +638,159 @@ func (u *ControlPlaneUpgrader) shouldSkipMachine(log logr.Logger, machine *clust
...
+
+	log := u.log.WithValues(
+		"machine", fmt.Sprintf("%s/%s", machine.Namespace, machine.Name),
+		"upgrade", u.upgradeID,
+	)
+	const deleteMachineInterval = 10 * time.Second
+	var err error
+
+	log.Info("Deleting machine")
+	err = wait.PollImmediate(deleteMachineInterval, u.machineTimeout, func() (bool, error) {
```
wfernandes: I think we can switch this to `wait.PollImmediateUntil`, so we can use the context created above instead of possibly blocking for 15 minutes.

pkg/upgrade/control_plane.go
```diff
@@ -673,18 +638,159 @@ func (u *ControlPlaneUpgrader) shouldSkipMachine(log logr.Logger, machine *clust
...
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		return errors.Wrapf(err, "timed out asking to delete machine %s/%s", machine.Namespace, machine.Name)
+	}
+
+	log.Info("Waiting for machine to be deleted")
+	err = wait.PollImmediate(deleteMachineInterval, u.machineTimeout, func() (bool, error) {
```
wfernandes: Same comment as above.

pkg/upgrade/control_plane.go
```diff
@@ -638,14 +600,17 @@ func (u *ControlPlaneUpgrader) shouldSkipMachine(log logr.Logger, machine *clust
 		return true
 	}
 
+	// Skip if this machine doesn't have NameBase annotation
 	baseName := machine.Annotations[AnnotationMachineNameBase]
 	if baseName == "" {
-		// should never happen
+		// should never happen because we reconcileMachineAnnotations for all
```
ncdc:
```suggestion
		// should never happen because we call reconcileMachineAnnotations for all
```

pkg/upgrade/control_plane.go
```diff
@@ -1251,18 +1373,46 @@ func (u *ControlPlaneUpgrader) waitForMatchingNode(ctx context.Context, rawProvi
...
+
+func (u *ControlPlaneUpgrader) hasMasterNodeLabel(nodeName string) bool {
+	log := u.log.WithValues("node", nodeName)
+	log.Info("Verifying node has master label", "label", LabelNodeRoleMaster)
+
+	node, err := u.targetKubernetesClient.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		log.Info("Node not found yet")
+		return false
+	} else if err != nil {
```
ncdc: nit: because you have a `return` inside the preceding `if` block, you don't need an `else` here

pkg/upgrade/control_plane_test.go
```diff
@@ -279,10 +288,11 @@ func TestShouldSkipMachine(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			u := &ControlPlaneUpgrader{
 				upgradeID: "abc",
+				log:       log.NullLogger{},
 			}
 
 			if e, a := tc.expectSkip, u.shouldSkipMachine(log.NullLogger{}, tc.machine); e != a {
-				t.Errorf("got %t, wanted %t", e, a)
+				t.Errorf("got %t, wanted %t", a, e)
```
ncdc: heh, for consistency, let's swap got/wanted instead of e/a 😄

pkg/upgrade/control_plane_test.go
```diff
@@ -392,3 +402,119 @@ kind: ClusterStatus
...
+					Annotations: map[string]string{
+						AnnotationUpgradeID:       upgradeID,
+						AnnotationMachineNameBase: "cp-0",
+					},
+				},
+			},
+			expected: true,
+		},
+		{
+			name: "machine is the original master of a cluster",
```
ncdc:
```suggestion
			name: "machine is an original control plane machine",
```

pkg/upgrade/control_plane.go
```diff
@@ -673,18 +638,175 @@ func (u *ControlPlaneUpgrader) shouldSkipMachine(log logr.Logger, machine *clust
...
+			// TODO make timeout the last arg, for consistency (or pass in a ctx?)
+			err = u.deleteEtcdMember(time.Minute*1, oldEtcdMemberID)
+			if err != nil {
+				return errors.Wrapf(err, "unable to delete old etcd member %s", oldEtcdMemberID)
+			}
+		}
+
+		// remove node from apiEndpoints in Kubeadm config map
+		log.Info("Removing machine from kubeadm ConfigMap")
+		err = wait.PollImmediate(deleteMachineInterval, u.machineTimeout, func() (bool, error) {
```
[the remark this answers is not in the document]
wfernandes: TBH this was originally polling so I just kept that as is.

pkg/upgrade/control_plane.go
```diff
@@ -673,18 +638,175 @@ func (u *ControlPlaneUpgrader) shouldSkipMachine(log logr.Logger, machine *clust
...
 
+// isHealthy verifies if a machine is healthy.
+// Conditions for health:
+// - the machine should have a ProviderID
+// - the machine should have a node associated with the ProviderID
+// - the node should be Ready
+// - the node should have the LabelNodeRoleMaster
+func (u *ControlPlaneUpgrader) isHealthy(machine *clusterv1.Machine) error {
+	ctx := context.TODO()
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
```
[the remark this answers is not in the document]
wfernandes: Yeah, I wasn't sure what time out to put here since these polling "can" effectively add 15 mins (by default) to the overall upgrade. I wasn't sure of adding other timeout configurations but I agree, I was a bit uncomfortable hardcoding timeouts. I'll switch to `u.machineTimeout` for now.

pkg/upgrade/control_plane.go
```diff
@@ -673,18 +638,175 @@ func (u *ControlPlaneUpgrader) shouldSkipMachine(log logr.Logger, machine *clust
...
+	return nil
+}
+
+// removeMachine removes the machine and its references.
+// It deletes and waits for the machine to be deleted.
+// If the machine has a ProviderID, it proceeds to remove the corresponding
+// etcd member and entry from the KubeadmConfigMap.
+func (u *ControlPlaneUpgrader) removeMachine(machine *clusterv1.Machine) error {
+	ctx := context.TODO()
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Minute)
```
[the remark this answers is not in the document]
wfernandes: Yup...agreed. 🙂

pkg/upgrade/control_plane.go
```diff
@@ -673,18 +638,175 @@ func (u *ControlPlaneUpgrader) shouldSkipMachine(log logr.Logger, machine *clust
...
+		} else {
+			log.Info("Replacement machine is healthy")
+		}
+	}
+
+	if len(ers) > 0 {
+		return errors.Wrap(ers, "errors when reconciling replacement machines")
+	}
+
+	return nil
```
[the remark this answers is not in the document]
wfernandes: Thanks...I was looking for this and ran across [some another type](https://github.com/kubernetes/apimachinery/blob/2373d029717c4d169463414a6127cd1d0d12680e/pkg/util/validation/field/errors.go#L227) and I was so confused.

pkg/upgrade/control_plane.go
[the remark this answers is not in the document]
ncdc: Yeah, I realized that after reading in more detail. I had removed other similar comments but I must have missed this one.

pkg/upgrade/control_plane.go
```diff
@@ -638,14 +606,17 @@ func (u *ControlPlaneUpgrader) shouldSkipMachine(log logr.Logger, machine *clust
 		return true
 	}
 
+	// Skip if this machine doesn't have NameBase annotation
 	baseName := machine.Annotations[AnnotationMachineNameBase]
 	if baseName == "" {
-		// should never happen
+		// should never happen because we call reconcileMachineAnnotations for all
+		// machines in the beginning.
 		log.Error(errors.New("machine is missing annotation "+AnnotationMachineNameBase), "unable to upgrade machine")
```
wfernandes: @ncdc Should this be logged as an `Error` or as `Info`? Even the previous condition which seems just as pertinent is only logged out as `Info`.

wfernandes:
/assign @chuckha 
/assign @vincepri 

Just trying to get more eyes on this PR. Let me know if y'all need context on what this PR is trying to fix.

ncdc:
@wfernandes do we need to examine all machines, including delete ones, at the beginning for any reason, e.g. when looking at the replacement machines? Would we ever want to wait for a busted replacement machine to be deleted before continuing?

wfernandes:
@ncdc That might be what's happening but it is really hard to tell what is happening from the unordered logs. 
Since we ignore the machines that are marked for deletion, it wouldn't show up in the list of machines to be processed. That is, if replacement machine `cp-0-upgrade0` is marked for deletion, it wouldn't show up in the list of machines. Then we would end up processing `cp-0` again and I would expect to see a failure in creating a replacement machine with the same name `cp-0-upgrade0` since it already exists.

pkg/upgrade/control_plane.go
chuckha: This should be an Info log.

My reasoning, and this does get somewhat subjective, is because this code handles this case and returns a valid value that the caller knows what to do with. Code execution is not stopped and therefore this isn't an error condition.

pkg/upgrade/control_plane.go
```diff
@@ -673,18 +646,170 @@ func (u *ControlPlaneUpgrader) shouldSkipMachine(log logr.Logger, machine *clust
...
+		)
+		log.Info("Found replacement machine for the in-process upgrade")
+
+		if err := u.isHealthy(m); err != nil {
+			log.Error(err, "Replacement machine is not healthy, going to remove all references of it")
+			if err := u.removeMachine(m); err != nil {
+				errs = append(errs, err)
+			}
+		} else {
+			log.Info("Replacement machine is healthy")
```
vincepri: You can use a `continue` in the first if block and spare an else here

pkg/upgrade/control_plane.go
```diff
@@ -267,8 +291,18 @@ func (u *ControlPlaneUpgrader) Upgrade() error {
 			Name:      replacementName,
 		}
 
-		if err := u.managementClusterClient.Get(context.TODO(), key, &replacement); err != nil {
-			return errors.Wrapf(err, "error getting machine %s", key.String())
+		err := u.managementClusterClient.Get(context.TODO(), key, &replacement)
```
vincepri: if err := ...; apierrors.IsNotFound(err) {
} else if err != nil {
...
}

pkg/upgrade/control_plane.go
```diff
@@ -267,8 +291,18 @@ func (u *ControlPlaneUpgrader) Upgrade() error {
...
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				// if there is a bad replacement machine in the beginning of the
+				// upgrade it is eventually deleted but the reference of the
+				// machine object is still in the slice of machines so just log
+				// this error and continue.
+				u.log.Error(err, "error getting machine", "machine", key.String())
```
vincepri: This could be also an `Info` message, including some of the information there is in the comment above

pkg/upgrade/control_plane.go
```diff
@@ -617,14 +573,28 @@ func (u *ControlPlaneUpgrader) reconcileMachineAnnotations(log logr.Logger, mach
...
 	return nil
 }
 
+func (u *ControlPlaneUpgrader) isReplacementMachine(machine *clusterv1.Machine) bool {
+	log := u.log.WithValues(
+		"machine", fmt.Sprintf("%s/%s", machine.Namespace, machine.Name),
+		"upgrade-id", u.upgradeID,
+	)
+	baseName := machine.Annotations[AnnotationMachineNameBase]
+	if baseName == "" {
```
vincepri: Would it be possible to put this validation before calling methods that need this annotation? That way we'd only have to check it once and we can error right away.

pkg/upgrade/control_plane.go
```diff
@@ -1251,18 +1376,47 @@ func (u *ControlPlaneUpgrader) waitForMatchingNode(ctx context.Context, rawProvi
...
+		log.Info("Node not found yet")
+		return false
+	}
+	if err != nil {
+		log.Error(err, "error getting node")
+		return false
+	}
+
+	_, ok := node.GetLabels()[LabelNodeRoleMaster]
+
```
vincepri: Nit: remove space

pkg/upgrade/control_plane.go
wfernandes: Thanks. I was leaning towards that as well.

pkg/upgrade/control_plane.go
wfernandes: TBH I don't think we need this check any more. I just refactored some previous logic into this method and left it here because I wanted to minimize unnecessary changes. For this case, if baseName was empty, the next condition check would be false which is the behavior we want anyways. 
Also this log would only serve as a distraction IMO.

pkg/upgrade/control_plane.go
```diff
@@ -673,18 +646,170 @@ func (u *ControlPlaneUpgrader) shouldSkipMachine(log logr.Logger, machine *clust
...
+	for _, m := range machines {
+		if !u.isReplacementMachine(m) {
+			continue
+		}
+
+		log := u.log.WithValues(
+			"machine", fmt.Sprintf("%s/%s", m.Namespace, m.Name),
+			"upgrade-id", u.upgradeID,
+		)
+		log.Info("Found replacement machine for the in-process upgrade")
```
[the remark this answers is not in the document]
wfernandes: It refers to the current upgrade run. I copied the terminology from line 622 in `shouldSkipMachine`.

wfernandes:
@ncdc @vincepri @chuckha I squashed all the commits into one. Let me know if there is any more I can do to improve this 🙂

vincepri:
LGTM

@ncdc ?

pkg/upgrade/control_plane.go
```diff
@@ -673,18 +632,171 @@ func (u *ControlPlaneUpgrader) shouldSkipMachine(log logr.Logger, machine *clust
...
+		return false, nil
+	}, ctx.Done())
+	if err != nil {
+		return errors.Wrapf(err, "timed out waiting for machine to be deleted %s/%s", machine.Namespace, machine.Name)
+	}
+
+	// Delete the machine and KubeadmConfigMap if there
+	// is a ProviderID for the machine.
+	// A replacement machine could've been created without a ProviderID
+	// assigned.
```
ncdc: Comment needs updating

pkg/upgrade/control_plane.go
```diff
@@ -673,18 +632,171 @@ func (u *ControlPlaneUpgrader) shouldSkipMachine(log logr.Logger, machine *clust
...
+	if machine.Spec.ProviderID != nil {
```
ncdc: How about checking `node` instead - it will have been set above.

PR #163, opened 2020-04-16
Remove only the OwnerReferences of kind "Machine"

### LINKED ISSUE #159: Bootstrap and InfraRef replacements do not honor OwnerRefs from the previous objects
Both the [KubeadmConfig](https://github.com/vmware/cluster-api-upgrade-tool/blob/a25fb714a2d2b0ea35302a9d80ad825f7321ce88/pkg/upgrade/control_plane.go#L893) and [InfraObject](https://github.com/vmware/cluster-api-upgrade-tool/blob/a25fb714a2d2b0ea35302a9d80ad825f7321ce88/pkg/upgrade/control_plane.go#L1045) have their owner references nil'd out when creating a replacement rather than just removing the old Machine reference.

In our case, these objects are managed and monitored as part of another higher-level object and so establishes an `OwnerRef` on them. After upgrading, the reference is dropped and the bootstrap and infra objects are filtered out for lacking that reference.

tvs:
This will preserve owner references that stem from other objects.

Fixes #159

### DIFF
--- a/pkg/upgrade/control_plane.go
+++ b/pkg/upgrade/control_plane.go
@@ -871,12 +871,14 @@ func generateReplacementMachineName(base, upgradeID string) string {
 	return machineName + machineSuffix
 }
 
-func getAllButMachineOwners(ownerRefs []metav1.OwnerReference) []metav1.OwnerReference {
+// filterOutMachineOwners filters out owner references of kind "Machine",
+// leaving owner references of other kinds. This allows the Machine controller
+// to adopt the objects after they're created, but allows us to preserve owner
+// refs of various other kinds
+func filterOutMachineOwners(ownerRefs []metav1.OwnerReference) []metav1.OwnerReference {
 	filtered := make([]metav1.OwnerReference, 0, len(ownerRefs))
 	for _, owner := range ownerRefs {
 		if owner.Kind == "Machine" {
-			// Skip Machine owners so that it will be adopted by the
-			// MachineController
 			continue
 		}
 		filtered = append(filtered, owner)
@@ -925,8 +927,7 @@ func (u *ControlPlaneUpgrader) updateBootstrapConfig(replacementKey ctrlclient.O
 	bootstrap.SetName(replacementKey.Name)
 	bootstrap.SetResourceVersion("")
 
-	// Remove only the Machine owner reference so we preserve other refs
-	ownerRefs := getAllButMachineOwners(bootstrap.GetOwnerReferences())
+	ownerRefs := filterOutMachineOwners(bootstrap.GetOwnerReferences())
 	bootstrap.SetOwnerReferences(ownerRefs)
 
 	// find node registration
@@ -1084,8 +1085,7 @@ func (u *ControlPlaneUpgrader) updateInfrastructureReference(replacementKey ctrl
 	infra.SetResourceVersion("")
 	infra.SetName(replacementKey.Name)
 
-	// Remove only the Machine owner reference so we preserve other refs
-	ownerRefs := getAllButMachineOwners(infra.GetOwnerReferences())
+	ownerRefs := filterOutMachineOwners(infra.GetOwnerReferences())
 	infra.SetOwnerReferences(ownerRefs)
 
 	unstructured.RemoveNestedField(infra.UnstructuredContent(), "spec", "providerID")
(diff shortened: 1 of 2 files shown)


pkg/upgrade/control_plane.go
```diff
@@ -859,6 +859,20 @@ func generateReplacementMachineName(base, upgradeID string) string {
 	return machineName + machineSuffix
 }
 
+func getAllButMachineOwners(ownerRefs []metav1.OwnerReference) []metav1.OwnerReference {
+	filtered := make([]metav1.OwnerReference, 0, len(ownerRefs))
+	for _, owner := range ownerRefs {
+		if owner.Kind == "Machine" {
+			// Skip Machine owners so that it will be adopted by the
+			// MachineController
```
imikushin, line 867: Nit: Is the purpose to filter out owners of kind Machine? If it is, this comment is unnecessary as it confuses the reader (me, at least) and makes them think if there's any other purpose of this very simple function.

pkg/upgrade/control_plane.go
```diff
@@ -890,7 +904,10 @@ func (u *ControlPlaneUpgrader) updateBootstrapConfig(replacementKey ctrlclient.O
 	// modify bootstrap config
 	bootstrap.SetName(replacementKey.Name)
 	bootstrap.SetResourceVersion("")
-	bootstrap.SetOwnerReferences(nil)
+
+	// Remove only the Machine owner reference so we preserve other refs
+	ownerRefs := getAllButMachineOwners(bootstrap.GetOwnerReferences())
+	bootstrap.SetOwnerReferences(ownerRefs)
```
imikushin, line 910: Is there a test for this change?

pkg/upgrade/control_plane.go
```diff
@@ -1042,7 +1059,11 @@ func (u *ControlPlaneUpgrader) updateInfrastructureReference(replacementKey ctrl
 	// prep the replacement
 	infra.SetResourceVersion("")
 	infra.SetName(replacementKey.Name)
-	infra.SetOwnerReferences(nil)
+
+	// Remove only the Machine owner reference so we preserve other refs
+	ownerRefs := getAllButMachineOwners(infra.GetOwnerReferences())
+	infra.SetOwnerReferences(ownerRefs)
```
imikushin, line 1065: Is there a test for this change?

pkg/upgrade/control_plane.go
tvs: Testing this change seems best suited for the integration test suite, which is currently broken. I've gone ahead and added a unit test for the filtering mechanism, though.

imikushin: I take it back: there are no unit test for `updateBootstrapConfig()` right now, so you can't update any existing unit test :(

Maybe let's add one for `getAllButMachineOwners()`, at least.

pkg/upgrade/control_plane.go
imikushin: Same here: taking it back for the lack of existing unit tests for `updateInfrastructureReference()`

pkg/upgrade/control_plane.go
```diff
@@ -859,6 +859,22 @@ func generateReplacementMachineName(base, upgradeID string) string {
...
 }
 
+// filterOutMachineOwners filters out owner references of kind "Machine",
+// leaving owner references of other kinds. This allows the Machine controller
+// to adopt the objects after they're created, but allows us to preserve owner
+// refs of various other kinds
+func filterOutMachineOwners(ownerRefs []metav1.OwnerReference) []metav1.OwnerReference {
+	filtered := make([]metav1.OwnerReference, 0, len(ownerRefs))
+	for _, owner := range ownerRefs {
+		if owner.Kind == "Machine" {
```
ncdc, line 869: @vincepri do you want to check api group as well?

vincepri: We could, for extra safety for sure

tvs: Good call. Gone ahead and added that.

pkg/upgrade/control_plane.go
```diff
@@ -859,6 +859,22 @@ func generateReplacementMachineName(base, upgradeID string) string {
...
 }
 
+// filterOutMachineOwners filters out owner references of kind "Machine",
+// leaving owner references of other kinds. This allows the Machine controller
+// to adopt the objects after they're created, but allows us to preserve owner
+// refs of various other kinds
+func filterOutMachineOwners(ownerRefs []metav1.OwnerReference) []metav1.OwnerReference {
+	filtered := make([]metav1.OwnerReference, 0, len(ownerRefs))
+	for _, owner := range ownerRefs {
+		if owner.APIVersion == clusterv1.GroupVersion.String() && owner.Kind == "Machine" {
```
ncdc, line 869: @vincepri this is matching on v1alpha2, which is probably ok since this tool is only for v1alpha2. WDYT?

PR #166, opened 2020-05-11
Added Machine patch option

tvs:
This patch is predominantly meant to remove custom labels from existing
Machines during the rollout process. For example: removing a label that
uniquely identified the Machine that was used to initialize the cluster.

### DIFF
--- a/pkg/upgrade/control_plane.go
+++ b/pkg/upgrade/control_plane.go
@@ -72,6 +72,7 @@ type ControlPlaneUpgrader struct {
 	upgradeID               string
 	oldNodeToEtcdMember     map[string]string
 	secretsUpdated          bool
+	machinePatch            jsonpatch.Patch
 	infrastructurePatch     jsonpatch.Patch
 	bootstrapPatch          jsonpatch.Patch
 	kubeadmConfigMapPatch   jsonpatch.Patch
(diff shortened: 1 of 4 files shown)


pkg/upgrade/control_plane.go
```diff
@@ -142,7 +143,13 @@ func NewControlPlaneUpgrader(log logr.Logger, config Config) (*ControlPlaneUpgra
 	infoMessage := fmt.Sprintf("Rerun with `--upgrade-id=%s` if this upgrade fails midway and you want to retry", config.UpgradeID)
 	log.Info(infoMessage)
 
-	var infrastructurePatch, bootstrapPatch, kubeadmConfigMapPatch jsonpatch.Patch
+	var machinePatch, infrastructurePatch, bootstrapPatch, kubeadmConfigMapPatch jsonpatch.Patch
+	if len(config.Patches.Machine) > 0 {
+		machinePatch, err = jsonpatch.DecodePatch([]byte(config.Patches.Machine))
+		if err != nil {
+			return nil, errors.Wrap(err, "error decoding infrastructure patch")
```
ncdc, line 150:
```suggestion
			return nil, errors.Wrap(err, "error decoding machine patch")
```

pkg/upgrade/control_plane.go
```diff
@@ -504,6 +512,15 @@ func (u *ControlPlaneUpgrader) updateMachine(replacementKey ctrlclient.ObjectKey
 		log.Info("New machine does not exist - need to create a new one")
 		replacementMachine = machine.DeepCopy()
 
+		// Patch first so we can overwrite any potential interferences with
```
ncdc, line 515: What sort of interferences?

pkg/upgrade/control_plane.go
```diff
@@ -1042,6 +1059,14 @@ func (u *ControlPlaneUpgrader) resourceExists(ref v1.ObjectReference) (*unstruct
 	return obj, nil
 }
 
+func patchMachine(machine *clusterv1.Machine, patch jsonpatch.Patch) (*clusterv1.Machine, error) {
```
ncdc, line 1062: I would remove this function and place the type assertion (cast) after you apply the patch above

ncdc:
@tvs will you be able to update this PR?

PR #170, opened 2020-06-30
Inclusive language update

tanglisha:
Updating language per VMware policy titled `Building Equitable and Inclusive Products`.


pkg/upgrade/control_plane.go
```diff
@@ -1427,7 +1427,7 @@ func (u *ControlPlaneUpgrader) UpdateProviderIDsToNodes() error {
 			id = providerID.ID()
 		} else {
 			u.log.Error(err, "failed to parse provider id", "id", node.Spec.ProviderID, "node", node.Name)
-			// unable to parse provider ID with whitelist of provider ID formats. Use original provider ID
+			// unable to parse provider ID with allowlist of provider ID formats. Use original provider ID
```
detiber, line 1430: I'm wondering if there is a better way to reword this, possibly:
```suggestion
			// unable to parse provider ID using known provider ID formats. Use original provider ID
```

